PR #23, opened 2019-06-25
Prepare for 1.0.0

rohanpm:
This library is now being taken into production use.
Bump the version to 1.0.0 to indicate that API compatibility is
now taken seriously.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ def get_requirements():
 
 setup(
     name="pubtools-pulplib",
-    version="0.3.0",
+    version="1.0.0",
     packages=find_packages(exclude=["tests"]),
     package_data={"pubtools.pulplib._impl.schema": ["*.yaml"]},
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -30,7 +30,7 @@ def get_requirements():
     long_description=get_long_description(),
     long_description_content_type="text/markdown",
     classifiers=[
-        "Development Status :: 3 - Alpha",
+        "Development Status :: 4 - Beta",
```
alexandrevicenzi: Should be 5 - Stable no? If we are moving to 1.0 it means that is stable

rohanpm: Not sure.  To me, 1.0.0 means that we are committing to doing semver-compatible version bumping from this point, e.g. any backwards-incompatible API changes will require the version to be bumped to 2.0.0.  But that's just about the API.  I don't have much evidence yet that the implementation actually works and is bug-free, so it feels premature to call it Stable.

PR #26, opened 2019-06-28
Support searching by regex, with minor refactors

rohanpm:
We have a use-case for searching repositories by regular expression,
so let's implement that.

Includes some backwards-compatible refactoring of the Criteria API.
Criteria and Matcher are now two separate concepts. This avoids
having to implement e.g. "with_field", "with_field_regex" and so
on methods for any type of matching we might want to do.


rohanpm:
@chandwanitulsi this change would make it possible for you to find repos based on notes.relative_url regex, as you were looking at today.

PR #28, opened 2019-07-17
Support uploading content to pulp

JayZ12138:
Implemented basic uploading functionality, which could upload a single file to pulp.


JayZ12138:
I didn't add the upload_one_file to fake.client, seems like it would just return None in the current implementation. I'm not sure if any exception handling is needed in upload_one_file function, if so, then I can do the same in the fake function.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,42 @@ def search_repository(self, criteria=None):
...
+    ):
+        """Including four steps uploading a file:
+            1. request upload id from pulp
+            2. upload contents to pulp
+            3. import the uploaded content to wanted repo
+            4. delete the reuqest id
+        """
+        upload_id = self._request_upload().result()["upload_id"]
+        offset = 0
+        with open(file_name) as f:
```
rohanpm: I think this is broken on Python 3, isn't it going to open the file as text and try decoding it using utf8 (or user's locale)?  Just tried it locally and it seems to be so.  I think you'll need to open with mode 'b' to read the binary data.

What I tried locally:

```
>>> open('tmp')
<_io.TextIOWrapper name='tmp' mode='r' encoding='UTF-8'>
>>> open('tmp', 'rb')
<_io.BufferedReader name='tmp'>
```

The first case will certainly crash or do the wrong thing when trying to upload e.g. an ISO or RPM file.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,42 @@ def search_repository(self, criteria=None):
...
+        unit_metadata=None,
+        override_config=None,
+    ):
+        """Including four steps uploading a file:
+            1. request upload id from pulp
+            2. upload contents to pulp
+            3. import the uploaded content to wanted repo
+            4. delete the reuqest id
+        """
+        upload_id = self._request_upload().result()["upload_id"]
```
rohanpm: You've written this all in blocking style, which means the current thread can't do anything else while waiting for the upload to be created, the import to finish, etc.  That means for example if someone needs to upload 10,000 files with this library, it'll be unnecessarily slow as it uploads and imports one file completely, then the next, then the next and so on, unable to handle any files concurrently.

Can you please ensure it's all in non-blocking style like the other methods?  It should be written in such a way that the call immediately returns a Future before any requests to Pulp have happened.

You probably want to use `f_map` and `f_flat_map` to achieve that.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,42 @@ def search_repository(self, criteria=None):
...
         )
 
+    def upload_one_file(
+        self,
+        file_name,
+        repo_id,
+        type_id,
+        unit_key,
+        unit_metadata=None,
+        override_config=None,
```
rohanpm: I'm wondering how much of this do you have a genuine use-case for yet?  If there are some bits which aren't needed currently, I recommend leaving them out instead.

I'm particularly looking at override_config, if you don't have a need to provide this, I think it'd be best to leave it out until it's needed.  It's hard to remove it once it's there, and unfortunately this API requires the caller to know many Pulp implementation details, such as the specific keys understood per override_config.

Actually, can we have some more complete discussion of the use-case first?  I think we need to consider, should we really add a generic upload API or should we rather add some APIs for specific content types we need to handle.

Consider that, if you add a generic upload API, then the users of this method all need to know Pulp implementation details like:

- the valid values for type_id
- what needs to be provided for unit_key
- what can be provided for unit_metadata, override_config

For example, the caller has to know - if I'm uploading an RPM then type_id should be "rpm", but if it's a source RPM then type_id should be "srpm".  I think it's within scope of this library to hide implementation details like that and let the caller interact with Pulp without having to know those kinds of things.

So, where are you planning on using this, should we maybe have an upload API for a specific content type instead?

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,42 @@ def search_repository(self, criteria=None):
...
+            repo_id,
+            upload_id,
+            unit_type_id=type_id,
+            unit_key=unit_key,
+            unit_metadata=unit_metadata,
+            override_config=override_config,
+        ).result()
+
+        self._delete_upload_request(upload_id).result()
+
```
rohanpm: Does the import task yield some information on the created units?  If so, might be better to return that.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,42 @@ def search_repository(self, criteria=None):
...
+        self._do_import(
+            repo_id,
+            upload_id,
+            unit_type_id=type_id,
+            unit_key=unit_key,
+            unit_metadata=unit_metadata,
+            override_config=override_config,
+        ).result()
+
+        self._delete_upload_request(upload_id).result()
```
rohanpm: I would rather suggest just fire-and-forget for deleting the upload request.  The caller doesn't actually need the upload request to be deleted before proceeding, and also if the deletion fails, the upload hasn't really failed.

In other words I think it's fine to just write:

```
# Queue deletion of upload request, but we don't wait for that to succeed
self._delete_upload_request(upload_id)
```

delete_upload_request probably should just log a warning if it fails though.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,42 @@ def search_repository(self, criteria=None):
...
+    def upload_one_file(
+        self,
+        file_name,
+        repo_id,
+        type_id,
+        unit_key,
+        unit_metadata=None,
+        override_config=None,
+    ):
+        """Including four steps uploading a file:
```
rohanpm: Can the doc string please be formatted so that it'll render properly, and document all the arguments in the same style as done for other methods?

Overall, I agree with the idea of having a method covering these 4 steps.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,42 @@ def search_repository(self, criteria=None):
             response_f, lambda data: self._handle_page(Repository, search, data)
         )
 
+    def upload_one_file(
```
rohanpm: Not certain currently, but I wonder if it might be better to put the public method onto Repository class instead.

I had similar decision to make for other methods.  For example, to publish a repository you call Repository.publish() instead of Client.publish(repo_id).  Either could work and I don't have a very strong sense of which way is better in general, but since the decision was already made to add publish and delete onto Repository class, I think upload probably makes sense there as well for the sake of consistency.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -302,3 +340,50 @@ def _delete_resource(self, resource_type, resource_id):
 
         LOG.debug("Queuing request to DELETE %s", url)
         return self._task_executor.submit(self._do_request, method="DELETE", url=url)
+
+    def _request_upload(self):
+        url = os.path.join(self._url, "pulp/api/v2/content/uploads/")
+
+        LOG.debug("Reuqesting upload id")
```
rohanpm: Typo in "Requesting"

tests/client/test_client.py
```diff
@@ -165,3 +165,68 @@ def test_get_missing(client, requests_mocker):
...
+            }
+        ],
+    }
+
+    tasks_report = [
+        {"task_id": "7744e2df-39b9-46f0-bb10-feffa2f7014b", "state": "finished"}
+    ]
+
+    somefile = tmpdir.join("some-file.txt")
+    somefile.write("I am the file to be uploaded")
```
rohanpm: If you used binary content here instead, the tests would probably catch that issue I mentioned before about not opening the file correctly.

e.g. somefile.write(b'some binary data:\x00\x01\x02')

tests/client/test_client.py
```diff
@@ -165,3 +165,68 @@ def test_get_missing(client, requests_mocker):
...
+
+    messages = caplog.messages
+    lines = [m.splitlines()[0] for m in messages]
+
+    # 6 reuqests should be made
+    assert requests_mocker.call_count == 6
+
+    # task's spwaned and completed
+    assert "Created Pulp task: 7744e2df-39b9-46f0-bb10-feffa2f7014b" in messages
+    assert "Pulp task completed: 7744e2df-39b9-46f0-bb10-feffa2f7014b" in messages
```
rohanpm: I think there's some verification missing here, like verifying the specifics of what was passed into import_upload and verifying that the upload was deleted.

rohanpm:
One more comment: as you're adding new public API, can you please also bump the minor version in setup.py in this commit?

i.e. if current version is x.y.z, because you've added new backwards-compatible API, per semver you should set next version to x.(y+1).z.

pubtools/pulplib/_impl/client/client.py
rohanpm: Thinking about it some more, a strong point in favor of Repository.upload is that the library most likely can figure out some things about the upload (at least the type_id and unit_key) based on the type of the Repository.

For example, for a FileRepository, the library knows that type_id must be "iso" and unit_key needs to have name, checksum and size. For a YumRepository, the library knows type_id must be one of ("rpm", "srpm", "yum_repo_metadata_file")... it can verify or maybe even autodetect these things.

tests/client/test_client.py
JayZ12138: got it.

pubtools/pulplib/_impl/client/client.py
JayZ12138: Okay.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -56,6 +56,9 @@ class Client(object):
     _REQUEST_THREADS = int(os.environ.get("PUBTOOLS_PULPLIB_REQUEST_THREADS", "4"))
     _PAGE_SIZE = int(os.environ.get("PUBTOOLS_PULPLIB_PAGE_SIZE", "2000"))
     _TASK_THROTTLE = int(os.environ.get("PUBTOOLS_PULPLIB_TASK_THROTTLE", "200"))
+    _PULP_CHUNK_SIZE = int(
+        os.environ.get("PUBTOOLS_PULPLIB_PULP_CHUNK_SIZE", 1024 * 1024)
```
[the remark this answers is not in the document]
JayZ12138: Makes sense

pubtools/pulplib/_impl/client/client.py
JayZ12138: import task will generate a task report, which is consumed by the task poller object and returns 5.0, which is the delay between poll attempts. I don't have an idea why this is returned.

pubtools/pulplib/_impl/client/client.py
JayZ12138: I have a question, can we upload as many files as we want with the same upload_request? If so, then I can change it to instead of upload_one_file, but to upload_files.

pubtools/pulplib/_impl/client/client.py
JayZ12138: Hm, I intended to upload the file for Maintainance mode in DELIVERY-7160, which is a json file. If just for this use case, I guess most options are not needed here.

pubtools/pulplib/_impl/client/client.py
JayZ12138: Based on the assumptions(one upload request could upload as many files as wanted and upload the same file could be uploaded by chunks in threads), I implemented it as:
request upload id, blocking => upload files in threads, every file upload is wrapped in a future, doesn't block each other, also upload the same file in threads as well, each part of the uploads doesn't block each other => waiting for upload to be done and then import, blocking => delete upload request, non-blocking, return a future. 
Is this Ok?

pubtools/pulplib/_impl/client/client.py
JayZ12138: Okay, I didn't see you put publish in Repository, I was confused since I didn't see how publish is tested, while the test coverage is 100%. 
I guess the name 'model' in model.repository confused me, since I thought it's more like a directory to store database related models, so I didn't check there. 

And yes, it's more reasonable to move the upload to Repository class.

pubtools/pulplib/_impl/client/client.py
rohanpm: One upload request can only upload one file.

pubtools/pulplib/_impl/client/client.py
rohanpm: OK, for that use-case, here is the API I'm thinking of:

```
class FileRepository(...):
  def upload_file(self, filename, relative_url=None):
```

- filename is path to a local file
- relative_url is the path & name which should be used in the remote repo.  It could default to os.path.basename(filename).

unit_key for "iso" units needs name, checksum and size.  checksum and size can be determined during the upload, and name comes from relative_url.  override_config and unit_metadata can be left out until we have some reason to provide them.

The upload method would be on FileRepository class and would be called upload_file, because the API is specific to "file" (or "iso") content and won't be exactly the same in the future for e.g. RPMs.

JayZ12138: Okay, I got it.

pubtools/pulplib/_impl/client/client.py
JayZ12138: Hm, then further change's needed.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - n/a
 
+## [1.2.0] - 2019-07-21
```
rohanpm: This isn't the way the changelog updates work, the idea is that you put your changes initially under "Unreleased".  Then, whoever is doing the releasing will move them under a new section with the right version number and date, at the time they do the release.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+###Added
```
rohanpm: Missing a bit of whitespace here.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -302,3 +333,40 @@ def _delete_resource(self, resource_type, resource_id):
...
+
+        LOG.debug("Importing contents to repo %s", repo_id)
+        return self._task_executor.submit(
+            self._do_request, method="POST", url=url, json=body
+        )
+
+    def _delete_upload_request(self, upload_id):
+        url = os.path.join(self._url, "pulp/api/v2/content/uploads/%s/" % upload_id)
+
+        LOG.debug("Deleting upload request")
```
rohanpm: May as well include the upload_id in the log here.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +181,36 @@ def search_repository(self, criteria=None):
             response_f, lambda data: self._handle_page(Repository, search, data)
         )
 
+    def _do_upload_file(self, upload_id, repo_id, filename):
+        tasks_f = f_return([])
+        offset = 0
+
+        def _do_next_part_upload(accumulated_tasks, upload_data):
+            data, upload_id, offset = upload_data
+            upload_task_f = self._do_upload(data, upload_id, offset)
```
rohanpm: This doesn't look right since do_upload doesn't return tasks.  Does it?

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +181,36 @@ def search_repository(self, criteria=None):
...
+                upload_task_f, lambda upload_task: accumulated_tasks + upload_task
+            )
+
+        # calculate size and checksum of the file during upload
+        with open(filename, "rb") as f:
+            checksum = hashlib.sha256()
+            size = 0
+            data = f.read(self._CHUNK_SIZE)
+            LOG.info("uploading %s to repo %s", filename, repo_id)
+            while data:
```
rohanpm: This is getting closer, but I think it's still doing too much in blocking style.  Specifically, it's reading entire file into memory before returning.

I don't think it will scale properly when there's many uploads.  A RHEL release for example may need to upload 40,000 files.  With the current approach here, if someone loops through all the files and calls upload for each, it will cause *all* content of *all* the files to be loaded in the memory of the current process, which probably won't work.

I'd expect that, to be scalable, the reading & checksumming of the file also has to be done in non-blocking style, i.e. you only read the next chunk once the current chunk has finished uploading.

Now that being said, I understand your current use-case isn't really needing to be scalable, so it's not a blocker to fix this - as long as the API isn't affected.  e.g. if we can make it more scalable later without changing the API, then it's acceptable to leave it like this (and probably file an issue to make it scalable).

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +181,36 @@ def search_repository(self, criteria=None):
...
+            return f_map(
+                upload_task_f, lambda upload_task: accumulated_tasks + upload_task
+            )
+
+        # calculate size and checksum of the file during upload
+        with open(filename, "rb") as f:
+            checksum = hashlib.sha256()
+            size = 0
+            data = f.read(self._CHUNK_SIZE)
+            LOG.info("uploading %s to repo %s", filename, repo_id)
```
rohanpm: Can we please have "Uploading" rather than "uploading"?

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +23,53 @@ class FileRepository(Repository):
     )
 
     mutable_urls = attr.ib(default=attr.Factory(lambda: ["PULP_MANIFEST"]), type=list)
+
+    def upload_file(self, filename, relative_url=None):
+        """Upload a file to this repository.
+
+        The upload operation includes 4 steps:
```
rohanpm: Actually, can you please *not* document it to this level of detail?  Sorry, I should have caught that on earlier review.

It's better if the doc doesn't specify exactly what steps are needed.  If you document that, then it can be argued we can't change it without breaking backwards-compatibility.  After all, you documented that it creates and deletes an upload, so it would be reasonable for someone to somehow rely on that.  That would block us from implementing behavior in this API such as: if file already exists in same repo with same name and checksum, skip the upload.

I think the docs should mostly focus on the postcondition of the method.  I'd put it like: Ensure this repository has a file at relative_url containing the content of filename.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +23,53 @@ class FileRepository(Repository):
...
+            4. Delete the upload request
+
+        Args:
+            filename (str)
+                Path to a local file to upload.
+
+            relative_url (str)
+                Path that should be used in remote repository
+
+                if omitted, filename will be used.
```
rohanpm: Can you please use full sentences with proper punctuation, so it'd be "if omitted" => "If omitted".

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +23,53 @@ class FileRepository(Repository):
...
+        Returns:
+            Future[:class:`~pubtools.pulplib.Task`]
+                A future which is resolved when publish succeeds.
```
rohanpm: This looks like copy-pasted and wrong since it's not a publish.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +23,53 @@ class FileRepository(Repository):
...
+
+                The future contains the task to delete upload request
```
rohanpm: Why do you return this task?  I don't think that's very useful to return.

Actually, deleting an upload doesn't even create a task - https://docs.pulpproject.org/dev-guide/integration/rest-api/content/upload.html#delete-an-upload-request.  So this can't be correct.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +23,53 @@ class FileRepository(Repository):
...
+
+        Raises:
+            DetachedException
+                If this instance is not attached to a Pulp client.
+        """
+        if not self._client:
+            raise DetachedException()
+
+        path, name = os.path.split(filename.rstrip("/"))
```
rohanpm: I don't understand the filename.rstrip ... if filename ends with "/" then it can't actually be a valid filename, can it?

For example, if I have a file "something.txt".  Why should I be able to upload_file("something.txt/") and have it automatically fixed to "something.txt"?  Shouldn't it just raise with an IO error since I gave an invalid filename?

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +23,53 @@ class FileRepository(Repository):
...
+        if not relative_url:
+            relative_url = path
```
rohanpm: I think this is not right, looks like you've set relative_url to only the dirname and omitted the basename.

It should allow using both.  The idea is that you might have a file named locally something like tmp.json but when you import it into the repo, you should be able to import it as e.g. "some/directory/my-file.json".  Both the directory and file names don't need to be same on Pulp as they are locally.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +23,53 @@ class FileRepository(Repository):
...
+
+        # request upload id and wait for it
+        upload_id = self._client._request_upload().result()["upload_id"]
+
+        # caculate hash/size of the file and upload it to pulp
+        upload_fs, checksum, size = self._client._do_upload_file(
```
rohanpm: It's an odd mix of blocking and non-blocking here.  What was the point of implementing the private method _do_upload_file as non-blocking if the only place you use it in public API is to immediately block on the result anyway?

tests/repository/test_upload.py
```diff
@@ -0,0 +1,92 @@
...
+    Repository,
+    FileRepository,
+    Task,
+    DetachedException,
+    TaskFailedException,
+)
+
+
+def test_upload_detached():
+    """publish raises if called on a detached repo"""
```
rohanpm: Copy-pasted and wrong.

tests/repository/test_upload.py
```diff
@@ -0,0 +1,92 @@
...
+        json=import_report,
+    )
+    requests_mocker.delete(
+        "https://pulp.example.com/pulp/api/v2/content/uploads/%s/" % upload_id, json=[]
+    )
+    requests_mocker.post(
+        "https://pulp.example.com/pulp/api/v2/tasks/search/", json=tasks_report
+    )
+
+    # delete's done, return []
```
rohanpm: It doesn't match what your docs say, which say that you'll return a "task" from deleting the upload.

tests/repository/test_upload.py
```diff
@@ -0,0 +1,92 @@
...
+    assert repo.upload_file(str(somefile)).result() == []
```
rohanpm: Looks like there's no testing of the case where relative_url is provided.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +23,53 @@ class FileRepository(Repository):
     )
 
     mutable_urls = attr.ib(default=attr.Factory(lambda: ["PULP_MANIFEST"]), type=list)
+
+    def upload_file(self, filename, relative_url=None):
```
rohanpm: It just occurred to me now that this API forces you into only being to upload content if you've already got it in a file on disk.

I'd expect that this may not even meet your current use-case all that well.  You're creating a JSON document, right?  Probably using Python "json" module?

So, with this API, don't you have to manipulate a dict in memory, use json to write it out to some file, then use this API to read it back from the same file again?  The data is written to disk and immediately read back?

Would it make more sense to let this take an IO object as well?  What I mean is:

```
def upload_file(self, file_obj, relative_url=None):
```

- if file_obj is a string, it's the name of a file to open.
- otherwise, if file_obj is a "file object" https://docs.python.org/3/library/io.html, data is read from that object

Could prevent you having to write content to disk just to ask this library to read it again.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +23,53 @@ class FileRepository(Repository):
...
+        Args:
+            filename (str)
+                Path to a local file to upload.
+
+            relative_url (str)
+                Path that should be used in remote repository
+
+                if omitted, filename will be used.
+        Returns:
+            Future[:class:`~pubtools.pulplib.Task`]
```
rohanpm: It's not returning this type, is it?

rohanpm:
> Now there's no public API in client now, is a fake one for _do_upload_file still needed?

What's needed is that, if you use the fake client to get a FileRepository instance, then the public upload_file method on that instance should be implemented and should do something reasonable according to the documentation.  It should probably at least crash if the file isn't readable.

CHANGELOG.md
JayZ12138: got it. 
I thought I would push the patch with a new tag so it would be released.

tests/repository/test_upload.py
JayZ12138: :(

tests/repository/test_upload.py
JayZ12138: yeah, I forget changing the docs.

tests/repository/test_upload.py
JayZ12138: I'll add it

pubtools/pulplib/_impl/model/repository/file.py
JayZ12138: yeah, makes sense. I'll make the change.

pubtools/pulplib/_impl/model/repository/file.py
JayZ12138: I wrote it to blocking since import can't be started until the file's uploaded to pulp, am I right?

pubtools/pulplib/_impl/client/client.py
JayZ12138: Right, it's a mistake here... 
it returns a request wrapped in a future, so I guess I actually don't need f_map here, as_completed should be enough for me?

JayZ12138: ah, I don't even need as_completed here, request_executor has done that for me.

CHANGELOG.md
rohanpm: Could you please not?  Designated code owners for this repo are myself and Rajul, I think part of the responsibilities should be acting as the gatekeeper for releases.  Especially for a library which is trying to follow semver, I think at least one of the owners should understand what's being released.

pubtools/pulplib/_impl/model/repository/file.py
rohanpm: Yeah, import can't be started until after the file's uploaded, but that doesn't mean you should use result() to block.  If you do that then it means the caller can't upload more than one file concurrently.

Here's a simplified pseudocode example.  In non-blocking style:

```
def upload_file(content, filename):
  upload_id_f = client._request_upload()
  # Once we have upload id then call upload_content
  upload_complete_f = f_flat_map(upload_id_f, lambda upload_id: client._upload_content(upload_id, content))
  # Once upload is complete then call import_content
  import_complete_f = f_flat_map(upload_complete_f, lambda _: client._import_content(upload_id_f.result(), ...))
  # Once import is complete then call delete_upload (but we don't care about the result)
  f_map(import_complete_f, lambda _: client._delete_upload(upload_id_f.result()))

  # Where are we now:
  # - we haven't necessarily *done* anything yet, just queued up the whole sequence
  # - hence the caller can immediately call us again to queue any more needed uploads as needed
  # - we'll return a future which will return whatever _import_content returned
  return import_complete_f
```

CHANGELOG.md
JayZ12138: Okay.

pubtools/pulplib/_impl/model/repository/file.py
JayZ12138: ah, I finally got what you mean. I didn't expect we could wrap things in such a way, it's really nice.

pubtools/pulplib/_impl/client/client.py
rohanpm: I think you don't actually care about the return value here, you just use the future to judge whether the operation completed.

In such cases I'm usually following the convention of "<verb>-ed",  e.g. "uploaded_f" is a future which is completed (or failed) when upload completes (or failed).

JayZ12138:
> Jay, could you consider writing an example for this too? It could help to ensure the API is right. Might help you develop and test it too.
> 
> For instance, in "examples" directory there could be `examples/upload-files` with behavior "upload all files in a directory to a given repo", run like e.g. `upload-files --repo-id my-repo ~/path/to/dir`
> 
> This, in combination with debug logging, could help with investigating how the blocking/non-blocking upload is working. For instance, if you had 10 files in the directory and you have blocking implementation you should be able to easily see that it's only handling one file at a time. If you have non-blocking implementation you should be able to get the example to upload multiple files at a time and it should be obviously faster.

Yeah, definitely.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +29,66 @@ class FileRepository(Repository):
...
 
     mutable_urls = attr.ib(default=attr.Factory(lambda: ["PULP_MANIFEST"]), type=list)
+
+    def upload_file(self, file_obj, relative_url=None):
+        """Upload a file to this repository.
+
+        Args:
+            file_obj (str, file object)
+                If it's a string, then it's the path of file to upload.
+                Else, it ought to be a file-like object
```
rohanpm: Might be nice to link to https://docs.python.org/3/glossary.html#term-file-object

rohanpm: Also, could you please stick with the style of using full sentences (so, end them with a full stop ".") ?

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +29,66 @@ class FileRepository(Repository):
...
+
+            relative_url (str)
+                Path that should be used in remote repository
+
+                If omitted, filename will be used.
```
rohanpm: There's no "filename" argument now.  I would rather word it as "the local name of the file will be used".

Also, I guess we'll have to crash if the user passed a file-like object without "name" attribute and without relative_url.  That's probably worth mentioning as well, like:

If `file_obj` is a file object without a `name` attribute, passing `relative_url` is mandatory.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +29,66 @@ class FileRepository(Repository):
...
+        is_path = isinstance(file_obj, str)
+        if is_path:
+            if not relative_url:
+                relative_url = file_obj
+            elif relative_url.endswith("/"):
+                _, name = os.path.split(file_obj)
+                relative_url = os.path.join(relative_url, name)
+        elif not is_path and (not relative_url or relative_url.endswith("/")):
+            msg = "Must provide complete relative_url if the file's not from disk"
+            LOG.exception(msg)
```
rohanpm: I don't think LOG.exception is correct here, as I understand it the point of exception() is that it'll dump traceback of the exception currently being handled - which can be also done via any other logger methods by including exc_info=1.

Since you're not yet handling an exception here, it doesn't seem to make sense.  And the message by itself doesn't contain enough info to know which piece of data caused the problem.  Probably just raising here is fine and letting whoever catches the exception worry about logging it.

Also, as a developer I would likely find useful a more detailed and complete message like:

```
"ValueError: %s is missing a `name` attribute and `relative_url` was not provided", file_obj
```

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +29,66 @@ class FileRepository(Repository):
...
+        )
+
+        f_map(
+            import_complete_f, lambda _: self._client._delete_upload_request(upload_id)
+        )
+
+        return import_complete_f
+
+    def _get_relative_url(self, file_obj, relative_url):
+        is_path = isinstance(file_obj, str)
```
rohanpm: This isn't the right isinstance check as mentioned elsewhere.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +29,66 @@ class FileRepository(Repository):
...
+        if is_path:
+            if not relative_url:
+                relative_url = file_obj
+            elif relative_url.endswith("/"):
```
rohanpm: Can you explain what this is about?  It seems to be a kind of feature:

If you specify "foo/" as relative_url, and the file is named "myfile.txt", then they're joined so the file is uploaded as "foo/myfile.txt" ?

This behavior isn't documented anywhere.  Can you please strip out the behavior, or include it in the documentation for the method?  Looks like we should have some more details on how relative_url is used.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +181,26 @@ def search_repository(self, criteria=None):
...
+        upload_fts = []
+
+        if isinstance(file_obj, str):
+            file_obj = open(file_obj, "rb")
+
+        checksum = hashlib.sha256()
+        size, offset = 0, 0
+        data = file_obj.read(self._CHUNK_SIZE)
+        LOG.info("Uploading %s to repo %s", file_obj, repo_id)
+        while data:
```
rohanpm: It seems like it's still doing too much up front in a blocking style.

Consider for example if you want to upload a 4GB ISO.  This will read the entire file into memory and enqueue 4,000 futures each holding a reference to a 1MB chunk.  So this method will have consumed 4GB of memory before returning any future.

I don't think it's right, it means that you can't upload any file with a size larger than the amount of memory you can allocate.  I don't think we should accept such a limit.

I think that you need to commit to making this either fully non-blocking (e.g. one future reads one chunk, next future reads another chunk ...), or if it's too complicated then make it fully blocking (a loop where you read a chunk, upload it, wait for that upload to finish, read next chunk, upload it, wait for that to finish ... etc).

pubtools/pulplib/_impl/client/client.py
```diff
@@ -302,3 +323,40 @@ def _delete_resource(self, resource_type, resource_id):
...
+            self._url, "pulp/api/v2/repositories/%s/actions/import_upload/" % repo_id
+        )
+
+        body = {
+            "unit_type_id": unit_type_id,
+            "upload_id": upload_id,
+            "unit_key": unit_key,
+        }
+
+        LOG.debug("Importing contents to repo %s", repo_id)
```
rohanpm: Putting upload_id in this message would be nice.

Otherwise if you're uploading dozens of files to the same repo, all the messages will be exactly the same and you'll have no way of matching up what was handled when.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +181,26 @@ def search_repository(self, criteria=None):
...
+
+        checksum = hashlib.sha256()
+        size, offset = 0, 0
+        data = file_obj.read(self._CHUNK_SIZE)
+        LOG.info("Uploading %s to repo %s", file_obj, repo_id)
+        while data:
+            checksum.update(data)
+            size += len(data)
+            upload_fts.append(self._do_upload(data, upload_id, offset))
+            offset += self._CHUNK_SIZE
```
rohanpm: Seems this ought to be `offset += len(data)`.

The code just a few lines above accepted that the size of data might be less than requested, but the code at this line would calculate an incorrect offset if that happened.

Per https://docs.python.org/3/library/io.html#io.RawIOBase.read

> Fewer than size bytes may be returned if the operating system call returns fewer than size bytes.

And per https://linux.die.net/man/2/read

> It is not an error if [the number of bytes read] is smaller than the number of bytes
> requested; this may happen [...] because read() was interrupted by a signal

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +181,26 @@ def search_repository(self, criteria=None):
...
+            data = file_obj.read(self._CHUNK_SIZE)
+        file_obj.close()
```
rohanpm: Hmm... not certain this is correct.  file_obj could be coming from the caller.  Are you sure they wanted their file closed?  Couldn't it have some undesirable side-effect for them...

Yeah, I'm pretty sure this is a bad idea if you simply consider the use-case: you have an open file and you want to upload the same thing to N repositories.  With this close here, the caller would have to open the same file N times for some reason.

I think you should avoid closing the file_obj when it was passed by the user.

tests/repository/test_upload.py
```diff
@@ -0,0 +1,118 @@
+import logging
```
rohanpm: Do we have the FakeController/Client update yet?  I'm not seeing a related test.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +29,66 @@ class FileRepository(Repository):
...
+        Returns:
+            Future[:class:`~pubtools.pulplib.Task`]
+                A future which is resolved when import succeeds.
+
+                The future contains the task to import uploaded content
+                to repository
+
+        Raises:
+            DetachedException
+                If this instance is not attached to a Pulp client.
```
rohanpm: Could you please use sphinx "versionadded" directive here to document when we've added this new API?

e.g.

```
  .. versionadded:: 1.2.0
```

examples/upload-files
```diff
@@ -0,0 +1,76 @@
...
+    logging.basicConfig(format="%(message)s", level=logging.INFO)
+
+    parser = ArgumentParser(description="Upload files to Repository")
+    parser.add_argument("--url", help="Pulp server URL")
+    parser.add_argument("--username", help="Pulp username")
+    parser.add_argument(
+        "--password", help="Pulp password (or set PULP_PASSWORD in env)"
+    )
+    parser.add_argument("--debug", action="store_true")
+    parser.add_argument("--repo-id", action="store")
```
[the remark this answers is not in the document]
JayZ12138: Yeah, I omitted this.

pubtools/pulplib/_impl/client/client.py
JayZ12138: wow, I'd say that's very hard to catch.

pubtools/pulplib/_impl/client/client.py
JayZ12138: hm, makes sense. 
Then I guess the right way is to make the upload_api only accepts file object, either it's an io.StringIO object or an opened file, then users could have more freedom.

tests/repository/test_upload.py
JayZ12138: Forget that. Doing it now.

pubtools/pulplib/_impl/model/repository/file.py
JayZ12138: Yeah, it's the feature you thought. 
In the previous commit, you mentioned that it should allow users to specify both path and name by passing relative_url, So I implemented like this, make it more flexible. 
Not sure if this is good or not.

pubtools/pulplib/_impl/client/client.py
JayZ12138: Okay, Then I think if users pass a string to the upload, then it should be closed.

pubtools/pulplib/_impl/model/repository/file.py
JayZ12138: A question here, if a file object has the `name` attribute, are we supposed to use that as remote path?

JayZ12138:
I'm working on the fake things now. I have a qurstion, feels like I should have a file.py under fake module?
OR should I just add the fake upload to client.py?

examples/upload-files
```diff
@@ -0,0 +1,77 @@
...
+                file = os.path.join(path, file)
+                log.debug("Uploading %s to repo %s in threads", file, repo_id)
+                uploads.append(repo.upload_file(file))
+    elif os.path.isfile(path):
+        log.debug("Uploading %s to repo %s", path, repo_id)
+        uploads.append(repo.upload_file(path))
+
+    for up in uploads:
+        result = up.result()
+        log.debug("Import task finished:\n", result)
```
rohanpm: Do you need some %s in the log message to ensure that result is logged?

examples/upload-files
```diff
@@ -0,0 +1,77 @@
...
+    parser.add_argument("--debug", action="store_true")
+    parser.add_argument("repo-id", action="store")
+    parser.add_argument("path", action="store", help="Path to a file or a directory")
+
+    p = parser.parse_args()
+
+    if not p.url:
+        parser.error("--url is required")
+
+    if not p.repo_id:
```
rohanpm: I think this is not actually needed?  Doesn't argparse already complain during parse_args now if repo_id is not provided?

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,28 @@ def search_repository(self, criteria=None):
...
+                    self._do_upload(data, upload_id, size),
+                    lambda _: do_next_upload(checksum, size + len(data)),
+                )
+            else:
+                # nothing more to upload, return checksum and size
+                if is_string:
+                    file_obj.close()
+                return f_return((checksum.hexdigest(), size))
+
+        is_string = isinstance(file_obj, six.string_types)
```
rohanpm: I think this is still not right.  I pointed to six.string_types doc but didn't expect you can just plug it in and it'll do what's needed, because it still doesn't work for bytes on Python3 (right?)

I thought you'll probably need to check both string_types and binary_type, if checking this way.

Alternatively, you can invert this check and decide that file_obj is a file object if and only if `'close' in dir(file_obj)` or similar.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,28 @@ def search_repository(self, criteria=None):
...
+        if is_string:
+            file_obj = open(file_obj, "rb")
+
+        LOG.info("Uploading %s to Pulp", file_obj)
```
rohanpm: This doesn't seem appropriate for an INFO log due to the fact that file_obj can be some object with no meaning to end-users.

For example I think this would end up with messages like these in Pub logs:

```
[INFO] Uploading <_io.BufferedReader name='/some/file'> to Pulp
```
Would it make more sense to log the name?  Or make this DEBUG if you wanted it for debugging purposes.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,28 @@ def search_repository(self, criteria=None):
...
+        return f_flat_map(f_return(), lambda _: do_next_upload(hashlib.sha256(), 0))
```
rohanpm: I was thinking at first that this was pointless, then I realized it's actually not because it means that if do_next_upload fails, that will result in a future which fails rather than this method raising an exception directly.

It could be considered desirable because it means there's just one method for closing file_obj in both cases (see earlier comment).

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,28 @@ def search_repository(self, criteria=None):
             response_f, lambda data: self._handle_page(Repository, search, data)
         )
 
+    def _do_upload_file(self, upload_id, file_obj):
+        def do_next_upload(checksum, size):
+            data = file_obj.read(self._CHUNK_SIZE)
```
rohanpm: Is there a problem here with how file_obj is managed?  Here, you assume file_obj is already a file object, but you may coerce it to a file object *after* do_next_upload is declared.  Does that actually work?  i.e. does this nested function capture file_obj at the point where do_next_upload is declared, in which case it's wrong, or does it somehow hold a reference to the outer scope which means it can see subsequent assignments to file_obj?

This isn't a rhetorical question as I'm not familiar with all details how closures work in Python.  If you're not sure, please test it.  Or even just refactor the code to avoid this - I often find a little callable helper class (replacing `do_next_upload` with a `class Upload` having a `__call__` method) ends up with more readable code.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -179,6 +182,28 @@ def search_repository(self, criteria=None):
...
+            if data:
+                checksum.update(data)
+                return f_flat_map(
+                    self._do_upload(data, upload_id, size),
+                    lambda _: do_next_upload(checksum, size + len(data)),
+                )
+            else:
+                # nothing more to upload, return checksum and size
+                if is_string:
+                    file_obj.close()
```
rohanpm: I think the file closing should be more robust, seems there are some error cases where file_obj.close may never be called.

I recommend trying a plain old future callback which will be invoked whether the future succeeds or fails.

So, the last few lines of _do_upload_file may be:

```
out = f_flat_map(f_return(), lambda _: do_next_upload(hashlib.sha256(), 0))
if is_string:
  # When upload completes, close the file we opened (whether or not it succeeded)
  out.add_done_callback(lambda _: file_obj.close())
return out
```

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +30,77 @@ class FileRepository(Repository):
...
     mutable_urls = attr.ib(default=attr.Factory(lambda: ["PULP_MANIFEST"]), type=list)
+
+    def upload_file(self, file_obj, relative_url=None):
+        """Upload a file to this repository.
+
+        Args:
+            file_obj (str, file object)
+                If it's a string, then it's the path of file to upload.
+                Else, it ought to be a file-like object, see:
+                https://docs.python.org/3/glossary.html#term-file-object
```
rohanpm: Could you please use a hyperlink instead of just adding the URL in the doc string?

What I mean is, the user should see "file-like object" and if they click on that, they should be directed to that doc.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +30,77 @@ class FileRepository(Repository):
...
+
+            relative_url (str)
+                Path that should be used in remote repository, can either
+                be a path to a directory or a path to a file, e.g:
+                - if relative_url is 'foo/bar' and file_obj has name 'f.txt',
+                  the result remote path wll be 'foo/bar/f.txt'.
```
rohanpm: result => resulting

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +30,77 @@ class FileRepository(Repository):
...
+        Args:
+            file_obj (str, file object)
+                If it's a string, then it's the path of file to upload.
+                Else, it ought to be a file-like object, see:
+                https://docs.python.org/3/glossary.html#term-file-object
+
+            relative_url (str)
+                Path that should be used in remote repository, can either
+                be a path to a directory or a path to a file, e.g:
+                - if relative_url is 'foo/bar' and file_obj has name 'f.txt',
```
rohanpm: This isn't correct, is it?  How would the code know in this case that you wanted foo/bar to be interpreted as a directory and not as a file?  It's perfectly valid to name a file 'foo/bar'...

I think in this example, you wanted: `foo/bar/` ?  You should probably make it explicit that a trailing slash is the method by which the code decides what to do.

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +30,77 @@ class FileRepository(Repository):
...
+        )
+
+        f_map(
+            import_complete_f, lambda _: self._client._delete_upload_request(upload_id)
+        )
+
+        return import_complete_f
+
+    def _get_relative_url(self, file_obj, relative_url):
+        is_path = isinstance(file_obj, six.string_types)
```
rohanpm: As mentioned elsewhere, I still don't think this is the right way to decide how to handle file_obj.

Also, you should probably encapsulate that in a helper function so you don't have to write the same code (fix the same bugs) twice.

pubtools/pulplib/_impl/model/repository/file.py
rohanpm: I think it's a decent feature.  Thinking about if we don't have this feature - the user passing a filename with trailing "/" doesn't really make sense (I don't think Pulp can store/publish a file like that), so we would be leaving the behavior undefined or adding some code to make it crash.

So let's keep it & just make sure it's documented/tested.

rohanpm:
> I'm working on the fake things now. I have a qurstion, feels like I should have a file.py under fake module?
> OR should I just add the fake upload to client.py?

For the similar `publish` case, I didn't add anything in the repository classes, I just ensured the fake client had the same internal API as the real client.  If that works here too then I'd recommend the same.

Bare minimum for the fake client would be for uploading to return a successful future.  I'd also recommend it tries to enforce the same requirements on its inputs as the real client does.  For example: if you pass the name of a file which can't be opened with the real client, that'll give a failing future, so the fake client should do the same.

examples/upload-files
JayZ12138: Yeah, forgot deleting it.

pubtools/pulplib/_impl/client/client.py
JayZ12138: Yeah, this confused me a bit. 
I thought bytes just wouldn't be used as a string anymore in python3.

pubtools/pulplib/_impl/client/client.py
JayZ12138: I think this is Okay, the function defined inside a function can access variables defined outside its body and it treats these variables as 'free variables', hold the references to them.
Here's a simple example, shows it keeps the reference to the variable. 
```
>>> def f():
...     def f1():
...             print(a)
...     a = 1
...     print(hex(id(a)))
...     return f1
... 
>>> p = f()
0x7f0c433974e0
>>> p.__closure__
(<cell at 0x7f0c35c93708: int object at 0x7f0c433974e0>,)
```

pubtools/pulplib/_impl/client/client.py
JayZ12138: Yeah, makes sense. I didn't think about unexpected failures... guess that's one reason why it's called 'unexpected'.

pubtools/pulplib/_impl/model/repository/file.py
JayZ12138: yep.

pubtools/pulplib/_impl/model/repository/file.py
JayZ12138: yes. I wanted /foo/bar/ :|

pubtools/pulplib/_impl/client/client.py
JayZ12138: I guess this means good?

pubtools/pulplib/_impl/client/client.py
rohanpm: Thanks Jay.  Repeated it locally on both python2 and python3, looks like it works fine.

pubtools/pulplib/_impl/fake/client.py
```diff
@@ -79,6 +82,35 @@ def get_repository(self, repository_id):
 
         return f_return(data[0])
 
+    def _do_upload_file(self, upload_id, file_obj, name):
+        # pylint: disable=unused-argument
+        out = f_return((hashlib.sha256().hexdigest(), random.randint(10, 1000)))
```
rohanpm: Is there a particular reason that you just return a static checksum and random size, instead of actually reading the file object?

I think it would be preferable to truly read the file object for a couple of reasons:

- so that it crashes if the file object can't actually be read - same as the real client would
- so that you get stable behavior with respect to the checksum & size, instead of using something random

Scaling isn't really a concern with the fake client, so I think you can feel free to implement that in a "simple" way, like just reading the data directly in one chunk.

pubtools/pulplib/_impl/fake/client.py
```diff
@@ -79,6 +82,35 @@ def get_repository(self, repository_id):
...
+    def _do_import(self, repo_id, upload_id, unit_type_id, unit_key):
+        # pylint: disable=unused-argument
+        task = Task(id=self._next_task_id(), completed=True, succeeded=True)
+
+        self._upload_history.append(Upload(repo_id, [task]))
+
+        return f_return([task])
+
+    def _delete_upload_request(self):
+        return f_return()  # pragma: no cover
```
rohanpm: Why no cover?  If this is never called then you don't even need to implement it.

I guess you really don't need to implement it, as deleting upload requests is an implementation detail not exposed in the API of the real client in any way, so there's no reason for the fake client to have any code for it.

tests/fake/test_fake_publish.py
```diff
@@ -63,3 +73,52 @@ def test_publish_absent_raises():
     exception = repo_copy2.publish().exception()
     assert isinstance(exception, PulpException)
     assert "repo1 not found" in str(exception)
+
+
+def test_can_upload(tmpdir):
```
rohanpm: Please don't add the upload test to the file named test_fake_publish, as publish and upload are different things.  In this project, I tried to keep test files smaller and more specific so we don't end up with what we see in many other projects, where each test file keeps growing indefinitely and it's tough to clean them up or to understand the relationship between tests and fixtures.

Can you please add that to a new file, test_fake_upload.py?  There's nothing wrong with having a test file starting with only 1-2 tests.

tests/fake/test_fake_publish.py
```diff
@@ -63,3 +73,52 @@ def test_publish_absent_raises():
...
+    repo1 = client.get_repository("repo1").result()
+
+    # If file's not found, Python 2 raises IOError and Python 3 raises
+    # FileNotFoundError. The latter one is not defined in Python 2.
+    if sys.version_info < (3,):
+        exception = IOError
+    else:
+        exception = FileNotFoundError
+    with pytest.raises(exception):
+        upload_f = repo1.upload_file("nonexistent_file")
```
rohanpm: I think this is actually pointing to a difference between the real and fake clients.

Question: if you use the real client and you attempt to upload a file which doesn't exist, does it:

* (A) - raise an exception, or
* (B) - return a future which raises an exception

I think it might be (B), but you've made the fake client do (A).

Can you please investigate and ensure they behave the same?

tests/repository/test_upload.py
```diff
@@ -0,0 +1,118 @@
...
+    requests_mocker.delete(
+        "https://pulp.example.com/pulp/api/v2/content/uploads/%s/" % upload_id, json=[]
+    )
+
+    assert repo.upload_file(str(somefile)).result() == [
+        Task(id="task1", succeeded=True, completed=True)
+    ]
+
+    # it' possible the delete has been called while doing assertion
+    assert requests_mocker.call_count == 5 or 6
```
rohanpm: This doesn't make sense.  Please consider the precedence of each part of this expression.

```
assert requests_mocker.call_count == 5 or 6
```

Is like writing:

```
assert( (requests_mocker.call_count == 5) or 6 )
```

And since 6 is always true, the assertion will always pass.

Try simplified version and note this assertion always passes even though the == part obviously is always false:

```
>>> assert 1 == 2 or 6
```

I think what you wanted was:

```
requests_mocker.call_count in [5, 6]
```

But actually, I think maybe you should rewrite this to wait for the deletion anyway during the test.  I mean, for example:

```
# Upload deletion is async and detached from our future - wait until it's done
for _ in range(0, 100):
  if requests_mocker.call_count == 6:
    break
  time.sleep(0.01)
assert requests_mocker.call_count == 6
```

And then you can actually verify in request_history that the deletion happened.

tests/repository/test_upload.py
```diff
@@ -0,0 +1,118 @@
...
+        u"name": str(somefile),
+        u"digest": u"fad3fc1e6d583b2003ec0a5273702ed8fcc2504271c87c40d9176467ebe218cb",
+        u"size": 29,
+    }
+    assert import_request["unit_key"] == import_unit_key
+
+    messages = caplog.messages
+    # task's spwaned and completed
+    assert "Created Pulp task: task1" in messages
+    assert "Pulp task completed: task1" in messages
```
rohanpm: As above, I'd like to suggest you wait for the upload deletion to happen and verify it too.

pubtools/pulplib/_impl/fake/client.py
```diff
@@ -13,6 +14,7 @@
 from .match import match_object
 
 Publish = namedtuple("Publish", ["repository", "tasks"])
+Upload = namedtuple("Upload", ["repo_id", "tasks"])
```
rohanpm: I'm not quite sure if this API with Upload and upload_history is complete enough.  I see a couple of issues with it:

Since it doesn't give you any info about what content was uploaded, all you can really check is whether the count of uploads is what you expected per repo.  So, for instance, if you had a class which was generating content and uploading them to Pulp repos, you couldn't use the fake client to check anything about the content uploaded, whether they were uploaded with correct filenames, etc.  It's really quite limited what you can do here.  This isn't a deal breaker since it could potentially have more attributes added later.

The bigger issue is - what about other content types in the future?  Right now, you've got upload_history and we only support file uploads.  Later we will support other kinds of uploads.  I'm thinking that it's quite possible different types of uploads ought to have different attributes collected in the history, so it may be a mistake to add one API which claims to support all kinds of uploads.

Can you please consider what to do about these issues?  Also, it's always an option to leave out that new history API altogether now and let someone add it when they have a use-case for it.

pubtools/pulplib/_impl/fake/client.py
```diff
@@ -79,6 +82,35 @@ def get_repository(self, repository_id):
...
+            "upload_id": "%s" % self._request_id,
+        }
+
+        return f_return(upload_request)
+
+    def _do_import(self, repo_id, upload_id, unit_type_id, unit_key):
+        # pylint: disable=unused-argument
+        task = Task(id=self._next_task_id(), completed=True, succeeded=True)
+
+        self._upload_history.append(Upload(repo_id, [task]))
```
rohanpm: Can you please make this fail if the repo doesn't exist, like a real client would?

pubtools/pulplib/_impl/fake/controller.py
```diff
@@ -79,3 +79,18 @@ def publish_history(self):
...
         return self.client._publish_history[:]
+
+    @property
+    def upload_history(self):
+        """A list of upload tasks triggered via this client.
+
+        Each element of this list is a named tuple with the following attributes,
+        in order:
+
+            ``repository``:
```
rohanpm: Documentation doesn't match reality.  The namedtuple you declared actually has a "repo_id" and not a "repository".

In this case, I think the documentation is better and the code should be fixed.  Publish returns a repository object and not a repo_id, why should Upload be inconsistent with that?  Can you please fix that to have a repository object as well?

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -20,3 +29,79 @@ class FileRepository(Repository):
...
+                  the resulting remote path wll be 'foo/bar/f.txt'.
+                - if relative_url is 'foo/bar/f.txt', no matter what the
+                  name of file_obj is, the remote path is 'foo/bar/f.txt'.
+
+                If omitted, the local name of the file will be used. Or,
+                if file_obj is a file object without a `name` attribute,
+                passing `relative_url` is mandatory.
+
+        Returns:
+            Future[:class:`~pubtools.pulplib.Task`]
```
rohanpm: Documentation maybe doesn't match reality again - isn't it returning a future with a *list of tasks*, not a single task?

pubtools/pulplib/_impl/fake/client.py
JayZ12138: hm, I felt I needed it since it's called in FileRepository.upload_file, though it's not consistent.
 
By consistent I mean, since we always return the import task, so the call of upload_file().result() can't promise _delete_upload_request is called, at least in the real cases. I'm not sure if it will be called in the fake client, I'll do some test

tests/fake/test_fake_publish.py
JayZ12138: whoops.

tests/fake/test_fake_publish.py
JayZ12138: make sense.

tests/repository/test_upload.py
JayZ12138: ah, overlooked this. 
Yeah, I thought about using time.sleep once, but just felt a little strange to sleep in the test.

pubtools/pulplib/_impl/fake/controller.py
JayZ12138: copy-pasted and wrong :(

pubtools/pulplib/_impl/model/repository/file.py
JayZ12138: copy-pasted and wrong again :(

pubtools/pulplib/_impl/fake/client.py
JayZ12138: Turns about it's never called. I removed it. But I don't know why it's never called.

pubtools/pulplib/_impl/fake/client.py
JayZ12138: yeah, Agree.

tests/fake/test_fake_publish.py
JayZ12138: Hm, actually, it just raises an exception directly, because open the file isn't wrapped in the a future, see code:
In file.py, upload_file calls _do_upload_file as:
`upload_complete_f = self._client._do_upload_file(upload_id, file_obj, name)`
in client.py, it opens the file first, then wrap the upload in future and return:
`is_file_object = "close" in dir(file_obj)
        if not is_file_object:
            file_obj = open(file_obj, "rb")

        LOG.info("Uploading %s to Pulp", name)
        upload_f = f_flat_map(f_return(), lambda _: do_next_upload(hashlib.sha256(), 0))`

So, this is not completely non-blocking, try to upload the file not existed will fail the whole process, is this OK?

pubtools/pulplib/_impl/fake/client.py
JayZ12138: Is this really necessary?
The use case for upload is to get repository first and then call repo.upload_file. If the repo doesn't exist, it should fail at get_repository, do we need to double-check here?

tests/fake/test_fake_publish.py
rohanpm: I am fine with either way as long as both the clients are consistent.

pubtools/pulplib/_impl/fake/client.py
rohanpm: I got this reply from Jay but now I can't find it in github for some reason:

> Is this really necessary?
> The use case for upload is to get repository first and then call repo.upload_file.
> If the repo doesn't exist, it should fail at get_repository, do we
> need to double-check here?

I would say that it is indeed still necessary since there can be any number of references to
the same repository.  I think there's already some autotests on the publish side which cover this.

For example you can have:

```
repo1 = client.get_repository('some-repo').result()
repo2 = client.get_repository('some-repo').result()

repo1.delete().result()

# and now try to upload to repo2
# expected behavior: it should fail!
repo2.upload_file(...)
```

pubtools/pulplib/_impl/fake/client.py
```diff
@@ -1,18 +1,20 @@
...
-from more_executors.futures import f_return, f_return_error
+from more_executors.futures import f_return, f_return_error, f_flat_map
 
 from pubtools.pulplib import Page, PulpException, Criteria, Task
 from .. import compat_attr as attr
 
 from .match import match_object
 
 Publish = namedtuple("Publish", ["repository", "tasks"])
+Upload = namedtuple("Upload", ["repository", "tasks", "unit_type_id", "unit_key"])
```
rohanpm: I think this is not too good either, exposing the unit_type_id and unit_key here encourages devs to write tests against Pulp2 implementation details where it's not necessary to do so.  upload_file is uploading a file "somehow" and the specifics of unit_type_id, unit_key are intentionally hidden, it's undermined by exposing them here like this.

If we are adding more attributes here I think they should try to be generic so as to expose as few Pulp2 implementation details as possible, while still meeting the requirements.  So, what requirement is being met by these?

I think it's probably: allowing the developer of a test to verify that correct content was uploaded to correct path.  Then I think it should be done more directly, like exposing "name" and ("sha256" or "content") attributes.

If you expose unit_key here, technically developers can look at unit_key["checksum"] to get the checksum.  But that kind of implementation detail - "On Pulp, ISO units have a 'checksum' in unit key" - is exactly what this library is meant to be hiding.

PR #30, opened 2019-08-02
Support uploading content to pulp

JayZ12138:
Pull request re-open for the one merged by mistake


JayZ12138:
about if fails the import task when repo doesn't exist in the fake client, I realized it's necessary after I left the comment and then deleted it, which caused confusion that it's in the email but can't find it in pull request anymore.

pubtools/pulplib/_impl/fake/client.py
```diff
@@ -79,6 +82,52 @@ def get_repository(self, repository_id):
...
+
+        if not is_file_obj:
+            out.add_done_callback(lambda _: file_obj.close())
+
+        return out
+
+    def _request_upload(self):
+        upload_request = {
+            "_href": "/pulp/api/v2/content/uploads/%s/" % self._request_id(),
+            "upload_id": "%s" % self._request_id,
```
rohanpm: This doesn't look right, as you didn't call self._request_id here, so the "%s" would be formatted as e.g. 

```
<function ... at 0x7fa2a7...>
```

Which leads me to question: why is everything apparently working even though you've got a junk upload_id here?

I guess it's because upload_id was ignored in _do_upload_file.  In terms of the public API, seems like this doesn't matter at all, but to protect against this kind of bug you might want to do some check of upload_id in that method.

pubtools/pulplib/_impl/fake/client.py
```diff
@@ -128,3 +177,6 @@ def _next_task_id(self):
         with self._lock:
             next_raw_id = self._uuidgen.randint(0, 2 ** 128)
         return str(uuid.UUID(int=next_raw_id))
+
+    def _request_id(self):
```
rohanpm: It's a little inconsistent with the naming of next_task_id.  Would this make more sense?

```suggestion
    def _next_upload_id(self):
```

pubtools/pulplib/_impl/fake/controller.py
```diff
@@ -79,3 +79,22 @@ def publish_history(self):
...
+    @property
+    def upload_history(self):
+        """A list of upload tasks triggered via this client.
+
+        Each element of this list is a named tuple with the following attributes,
+        in order:
+
+            ``repository``:
+                :class:`~pubtools.pulplib.Repository` for which upload was triggered
+            ``task``:
```
rohanpm:
```suggestion
            ``tasks``:
```

pubtools/pulplib/_impl/fake/controller.py
```diff
@@ -79,3 +79,22 @@ def publish_history(self):
...
+                :class:`~pubtools.pulplib.Task` generated as a result
```
rohanpm:
```suggestion
                list of :class:`~pubtools.pulplib.Task` generated as a result
```

tests/fake/test_fake_upload.py
```diff
@@ -0,0 +1,78 @@
...
+    assert tasks[0].succeeded
+
+    # The change should be reflected in the controller's upload history
+    history = controller.upload_history
+
+    digest = "fad3fc1e6d583b2003ec0a5273702ed8fcc2504271c87c40d9176467ebe218cb"
+    assert len(history) == 1
+    assert history[0].repository == repo1
+    assert history[0].tasks == tasks
+    assert history[0].name == str(somefile)
```
rohanpm: Shouldn't it actually be just "some-file.txt" - I mean just the basename and not the entire path with leading tmpdir?

```suggestion
    assert history[0].name == "some-file.txt"
```

If tests don't pass with that change then something might be wrong.

tests/repository/test_upload.py
```diff
@@ -0,0 +1,121 @@
...
+    requests_mocker.delete(
+        "https://pulp.example.com/pulp/api/v2/content/uploads/%s/" % upload_id, json=[]
+    )
+
+    assert repo.upload_file(str(somefile)).result() == [
+        Task(id="task1", succeeded=True, completed=True)
+    ]
+
+    # sleep for .1 second to let the whole process finish.
+    time.sleep(0.1)
```
rohanpm: I think you've got to loop, this can be too flaky.  You might not be able to reproduce a failure locally, but when this is running on a VM or container in CI which could be shared by many other jobs, it's really possible that 0.1 seconds sometimes isn't enough for the other thread to be scheduled.

I usually would do something like: loop up to 1000 iters with 0.01 seconds each, so test is fast when passing and waits up to 10 seconds before failing.

If there are several tests which would need to do that, then it'd be time for a helper function.

tests/repository/test_upload.py
```diff
@@ -0,0 +1,121 @@
...
+    import_request = requests_mocker.request_history[3].json()
+    import_unit_key = {
+        u"name": str(somefile),
+        u"digest": u"fad3fc1e6d583b2003ec0a5273702ed8fcc2504271c87c40d9176467ebe218cb",
+        u"size": 29,
+    }
+    assert import_request["unit_key"] == import_unit_key
+
+    messages = caplog.messages
+    # task's spwaned and completed
```
rohanpm:
```suggestion
    # task's spawned and completed
```

tests/fake/test_fake_upload.py
JayZ12138: Hm, I feel this should be the full path. 
The 'name' here is actually unit_key["name"], which is relative_url passed to import, which will be used as the remote path, so I think the full path should be used. 
And the behavior in fake client should be consistent.

rohanpm: OK, we might not have had the same idea over how the API should actually work then.

IMHO, in the case where the caller didn't provide relative_url during upload, the full path to the file locally should NOT be used to set relative_url, only the basename should be used.  That's because the full path is almost certain to be meaningless.

For example if the user prepared some content to push under: /mnt/staging/my-push-123/repo1/ISO/rhel-boot-dvd.iso

Then, while it could make sense to default the filename to "rhel-boot-dvd.iso" during upload, I don't think it could make sense to include that whole /mnt/staging/... path.

In practice, relative_url 99% of the time would be only a basename I think.

examples/upload-files
```diff
@@ -0,0 +1,80 @@
...
+log = logging.getLogger("upload")
+
+
+def upload(client, path, repo_id):
+    repo = client.get_repository(repo_id).result()
+
+    uploads = []
+    if os.path.isdir(path):
+        for file in os.listdir(path):
+            if os.path.isfile(file):
```
rohanpm: I just gave this a try locally and I don't think it works (or it may only work in the case where path is ".")

os.listdir returns filenames relative to the directory being listed.  If path isn't the current working directory then those files won't be accessible by the relative paths.

e.g. if path is /some/directory and /some/directory/file1, /some/directory/file2 exist, then "file" would be e.g. "file1", os.path.isfile("file1") will return False unless the current working directory happens to be /some/directory.

Looks like the os.path.join should be moved to before this isfile check and then it'd work.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -302,3 +326,40 @@ def _delete_resource(self, resource_type, resource_id):
...
+
+    def _request_upload(self):
+        url = os.path.join(self._url, "pulp/api/v2/content/uploads/")
+
+        LOG.debug("Requesting upload id")
+        return self._request_executor.submit(self._do_request, method="POST", url=url)
+
+    def _do_upload(self, data, upload_id, offset):
+        url = os.path.join(
+            self._url, "pulp/api/v2/contents/%s/%s" % (upload_id, offset)
```
rohanpm: This seems like the wrong URL.  See the docs at: https://docs.pulpproject.org/dev-guide/integration/rest-api/content/upload.html#upload-bits

```suggestion
            self._url, "pulp/api/v2/content/uploads/%s/%s" % (upload_id, offset)
```

examples/upload-files
```diff
@@ -0,0 +1,81 @@
...
+    parser = ArgumentParser(description="Upload files to Repository")
+    parser.add_argument("--url", help="Pulp server URL")
+    parser.add_argument("--repo-id", action="store")
+    parser.add_argument("--path", action="store", help="Path to a file or a directory")
+    parser.add_argument("--username", help="Pulp username")
+    parser.add_argument(
+        "--password", help="Pulp password (or set PULP_PASSWORD in env)"
+    )
+    parser.add_argument("--debug", action="store_true")
+    parser.add_argument("--verify", default=False, action="store_true")
```
rohanpm: Can you please invert this so that it verifies by default?

If in doubt, copy the option name from an existing command.  e.g. curl uses -k, --insecure.

JayZ12138:
> The example is now working for me locally.
> 
> I think this is close to merge, but I would rather not have all the intermediate commits in history, do you want to rebase and clean up the git history yourself or should we do it at merge time?

Yeah. I guess 'squash and merge' while doing merge will work?

examples/upload-files
JayZ12138: hm, it's weird that I have already done the changes on my side and pushed, but it doesn't change in this pull request. Not sure why.

JayZ12138:
Since the change in example/uploads-files didn't show, I'm closing it and reopen a new one.

PR #35, opened 2019-08-13
Introduce the Repository.is_temporary attribute

rohanpm:
This attribute tells us whether a repository should be considered
"temporary" (i.e. it's created to support execution of some
workflow and should be deleted once that's complete).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- n/a
+### Added
+- Introduced ``Repository.is_temporary`` attribute
 
 ## [1.2.1] - 2019-08-12
 
--- a/pubtools/pulplib/_impl/model/repository/base.py
+++ b/pubtools/pulplib/_impl/model/repository/base.py
@@ -116,6 +116,19 @@ class Repository(PulpObject):
     """True if this is a sigstore repository, used for container image manifest
     signatures."""
 
+    is_temporary = pulp_attrib(
+        default=False, type=bool, pulp_field="notes.pub_temp_repo"
+    )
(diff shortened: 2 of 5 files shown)


rohanpm:
@rajulkumar this is in response to the autotest you added here: https://github.com/release-engineering/pubtools-pulp/blob/master/tests/test_garbage_collect.py, because I see you used private API there, which is bad.

I consider it in scope of pubtools-pulp to encapsulate all logic like "notes.pub_temp_repo exists and is used to mark repo as a temporary", instead of having other projects needing to know that.

rajulkumar:
Agree that private API shouldn't be used. However, was wondering whether it's fine to add this new attribute to the Repository class based on the "notes" as there are a lot more values in "notes" that might be required in future. Also, this attribute is only required for tests and doesn't affect the actual code.

rohanpm:
In my view, it's in scope for this library to encapsulate *all* the notes we might want to use.  The schema in here should be the definitive source when someone is wanting to investigate which notes might be found on our Pulp servers and what do they mean.

About it only being needed for tests - I guess you mean because the fake controller/client can only search on defined fields (via pulp_attrib)?

I think this shows I've made a bit of a design mistake with the search functionality, and I have a plan how to fix it.  Right now, to find these repos you have to search for "notes.pub_temp_repo", so you have to know that this note exists and it's not abstracted fully by this new is_temporary attribute.  Now I'm thinking you should be able to search by using the field names defined in this library, i.e. search for a repo by field "is_temporary".  That way, both the real and test code in pubtools-pulp will only be dealing with "is_temporary" field, and "notes.pub_temp_repo" will genuinely be just an implementation detail of this library.

So, I plan to work on a PR which implements that.  We could wait until that's ready and have another look at this if you want to get a better idea what I mean.

rajulkumar:
I see what you mean. For the idea of encapsulation this change makes sense.

PR #36, opened 2019-08-15
 Allow searching on model fields rather than (only) raw Pulp fields

rohanpm:
Allows to search directly for e.g. "is_temporary", rather than having to know the
underlying field in which that data is stored, "notes.pub_temp_repo".

This reduces dependency on Pulp implementation details.

### DIFF
--- a/tests/client/test_search_repo_conversions.py
+++ b/tests/client/test_search_repo_conversions.py
@@ -1,0 +1,61 @@
+import pytest
+
+from pubtools.pulplib import Criteria, Matcher, Repository
+
+from pubtools.pulplib._impl.client.search import filters_for_criteria
+
+
+def test_eng_product_in():
+    """eng_product is mapped correctly"""
+    crit = Criteria.with_field_in("eng_product_id", [12, 34, 56])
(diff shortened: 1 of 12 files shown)


rohanpm:
@rajulkumar this is supposed to help you completely scrap the need to embed "notes.pub_temp_repo" anywhere, so it's fully encapsulated by this library.

tests/client/test_search_repo_conversions.py
```diff
@@ -0,0 +1,61 @@
+import pytest
+
+from pubtools.pulplib import Criteria, Matcher, Repository
+
+from pubtools.pulplib._impl.client.search import filters_for_criteria
+
+
+def test_eng_product_in():
```
rajulkumar: minor: 'test_eng_product_id'

rohanpm: Actually I did mean "in" because the test is doing a search with "in" operator, though I see how it's ambiguous.

rajulkumar: I see what you mean. It's called eng_product_id, so looked like a typo.

PR #37, opened 2019-08-15
Support get maintenance status of pulp repositories

JayZ12138:


pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,6 +181,15 @@ def search_repository(self, criteria=None):
             response_f, lambda data: self._handle_page(Repository, search, data)
         )
 
+    def get_maintenance_status(self):
+        """Get maintenance status from maintenance repository
+
+        Returns:
+            Future[dict]
```
rohanpm: Could you please make this a proper object (attrs-based class, perhaps a PulpObject subclass) with a documented and enforced interface?

I don't want to be returning raw dicts here, it means anyone using this API needs to somehow know what's in the dict and how to interpret it.  It should rather be the job of this library to interpret the raw data from Pulp.

You should probably also add a new schema for the file.  As far as I know, nobody wrote a proper reference for the format of this file yet.  This library seems a reasonable place for it to live.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -363,3 +372,8 @@ def _delete_upload_request(self, upload_id):
 
         LOG.debug("Deleting upload request %s", upload_id)
         return self._request_executor.submit(self._do_request, method="DELETE", url=url)
+
+    def _do_get_maintenance(self):
+        url = os.path.join(self._url, "pulp/isos/redhat-maintenance/repos.json")
+
+        return self._request_executor.submit(self._do_request, method="GET", url=url)
```
rohanpm: What happens if file is missing (404)?  In my opinion we should be returning some default indicating that the system is not under maintenance.

pubtools/pulplib/_impl/fake/controller.py
```diff
@@ -98,3 +100,19 @@ def upload_history(self):
                 checksum of the file uploaded
         """
         return self.client._upload_history[:]
+
+    def update_maintenance_file(self):
```
rohanpm: Why is the behavior hardcoded like this?  I suggest the way it should work is:

- as mentioned in earlier comment, there should be some proper object (not raw dict) modelling the maintenance status
- then you could have set_maintenance(x) which simply stores x as the new maintenance status

Also, this is a public method and should be documented.

examples/get-maintenance-json
```diff
@@ -0,0 +1,57 @@
...
+
+from pubtools.pulplib import Client
+
+log = logging.getLogger("get-maintenance")
+
+
+def get_maintenance(client):
+
+    maintenance = client.get_maintenance_status().result()
+    log.info("maintenance status: %s", maintenance)
```
rohanpm: I think this example is probably a bit *too* trivial.  Can it be something more realistic?

Like instead of a raw dump, tell us:

- which repo(s) are in maintenance mode
- who set them in maintenance mode / the comment

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,6 +181,15 @@ def search_repository(self, criteria=None):
             response_f, lambda data: self._handle_page(Repository, search, data)
         )
 
+    def get_maintenance_status(self):
+        """Get maintenance status from maintenance repository
```
rohanpm: The docs here should have a bit of explanation about what "maintenance status" is.

Also, you don't need to say it's from "maintenance repository" as this should be just an implementation detail, ideally hidden.

pubtools/pulplib/_impl/client/client.py
JayZ12138: Hm, from JIRA description, seems like there will always be a repos.json file, if some repos in maintenance mode, then there will be entires in repos:{repo_id:{...}}, if not, then just no entires there, but the file is always there. 
I'm thinking this feature isn't that useful?

pubtools/pulplib/_impl/fake/controller.py
JayZ12138: set_maintenance is supposed to be done in pubtools.pulp, I'm not sure how to do it here?

pubtools/pulplib/_impl/client/client.py
rohanpm: You mean it wouldn't be useful because in practice you always expect the file to be present on the real system?

I think it is in fact still useful to tolerate its absence because it lowers the bar for getting a working environment for use with this library (and tools built on top of it).  For example, imagine that there is a Pub feature built on this library which somehow displays or reacts to maintenance mode, e.g. refuses to run a task if maintenance mode is set, and imagine you are trying to develop with Pub locally to make unrelated changes to that task.

If you say, "it's only valid to use this library against a Pulp server which has the maintenance repo and the repos.json file", you just made life harder for that developer.  Now the developer can't simply run Pub and connect it with some clean local Pulp dev env; they also have to ensure that Pulp server has the repo and the file, or else they immediately hit a wall.

The same applies for autotests - if you require that this file must always be there or this library will crash, then you also forced anyone mocking out the Pulp server at the HTTP level to know about this URL and know how to provide valid data for it (instead of just e.g. defaulting to 404 for unknown URLs, which is reasonable).

So if you make this code only work when this file is present, you just made it harder to set up a working test or development environment.

pubtools/pulplib/_impl/fake/controller.py
rohanpm: That's actually a good point as well and something I missed:

I think you should also have the logic for setting maintenance mode belonging to this library.  In pubtools-pulp, it would mainly be just calling a method or two to get/set the mode according to arguments.  That would ensure all the details about e.g. which repo contains that file, what's the format of the file, are all encapsulated in this library and not presented elsewhere.

midnightercz:
why there's no description?

pubtools/pulplib/_impl/fake/controller.py
midnightercz: why is this method implemented only in fake client?

midnightercz:
Thinking about this more, I believe this shouldn't be part of this repository. How is maintenance modes related to pulp anyway. It's our customization that has nothing to do with pulp and it's functionality. I think it makes sense to this whole thing into pubtoolslib-pulp

rohanpm:
It's in the intended scope of this library to also cover our customized usage of Pulp.  That's why it already has plenty of things related to our customizations, e.g. the documented schemas which cover notes such as notes.eng_product.

I think it's best to try to encapsulate those customizations here rather than let their implementation details be spread out over every piece of code which needs to deal with them.

pubtools/pulplib/_impl/fake/controller.py
JayZ12138: Hm, this is in the controller, not client.

JayZ12138:
> why there's no description?

I thought it's a quite straightforward PR, the title says everything, so didn't add a description.

pubtools/pulplib/_impl/client/client.py
rohanpm: I've just had the realization that this is going to use the same requests session as used for the Pulp API, which means the same "auth".  This is a bit weird because the auth is only required for requests to the API, and requests for content like this should not need any authentication.

I *guess* that this shouldn't cause a problem and it's OK to pass the auth even when not required, but could you please check if you haven't already tested this against the real Pulp servers with auth?

JayZ12138: Yeah, I checked on the rhsm-pulp-dev, it works fine, at least for getting the file.

JayZ12138:
Hi @rohanpm,

This is still in progress, but I want to check if the idea by far is OK or not. Tests and documentation haven't been added. 

There are 3 known issues existed, I described them in the commit message.

rohanpm:
Why do you say that upload API can't overwrite the file in place?

It should be able to.  It's the desired behavior that putting an ISO of a given name into a repo, would overwrite an ISO with the same name.

There was a regression bug DELIVERY-7037 which prevented this from working correctly, maybe have a look at that issue and make sure the fix is present on whatever Pulp server you've tested against?  I think in this case, we could target Pulp servers with a fix for that bug rather than complicating our code here to work around it.

At some point we should have unassociate API here, but I'd like us to put more effort into it and probably introduce some classes to model "unit" objects properly.

JayZ12138:
> Why do you say that upload API can't overwrite the file in place?
> 
> It should be able to. It's the desired behavior that putting an ISO of a given name into a repo, would overwrite an ISO with the same name.
> 
> There was a regression bug DELIVERY-7037 which prevented this from working correctly, maybe have a look at that issue and make sure the fix is present on whatever Pulp server you've tested against? I think in this case, we could target Pulp servers with a fix for that bug rather than complicating our code here to work around it.
> 
> At some point we should have unassociate API here, but I'd like us to put more effort into it and probably introduce some classes to model "unit" objects properly.

I tested against rhsm-pulp-dev, with latest dev-env setup. 
When I tried to upload the file with the same name to it, nothing happened. Then I checked the import task, it shows the error 'File Exists'. I didn't find any official documentation indicates how to overwrite an existed file or says it overwrites by default. I'm thinking if I made any mistakes, such as upload the exact same file to the same repo? I'll double-check tomorrow.

rohanpm:
OK, there's a chance you see something different from DELIVERY-7037 because we normally upload an ISO first into one repo (all-iso-content) and then associate it with a target repo.  Here, you instead use the more direct approach of uploading & importing straight to the target repo.  That's fine but might be triggering some different behavior.

I'm not familiar with that "File Exists" response, if you can confirm it then maybe you can paste the entire resulting Pulp task JSON here for us to have a look at?

JayZ12138:
> OK, there's a chance you see something different from DELIVERY-7037 because we normally upload an ISO first into one repo (all-iso-content) and then associate it with a target repo. Here, you instead use the more direct approach of uploading & importing straight to the target repo. That's fine but might be triggering some different behavior.
> 
> I'm not familiar with that "File Exists" response, if you can confirm it then maybe you can paste the entire resulting Pulp task JSON here for us to have a look at?

Okay, I just double-checked. 
The problem is not related to import task directly. The root cause is when I upload a file with the same name of a file that already existed in the repo, it doesn't overwrite the existed one, but ends up with two files with same names and then fails publish because the file already exists.
Here's 
[upload-publish.log](https://github.com/release-engineering/pubtools-pulplib/files/3525523/upload-publish.log)

JayZ12138:
I created a new PR without the unassociate code, this could be closed if we can confirm the upload is a bug and has been fixed.

rohanpm:
Jay, would you like to drop this one now as obsoleted by https://github.com/release-engineering/pubtools-pulplib/pull/41 ?

JayZ12138:
> Jay, would you like to drop this one now as obsoleted by #41 ?

yes.

PR #40, opened 2019-08-20
match_field_exists in fake.match.py to return false for fields with None

### LINKED ISSUE #39: Matcher.exists returns true when model attribute is None while using Fake client
#### Summary

Matcher.exists() return true if a model attribute is used in the search criteria and that attribute is None.
This applies only to the Fake client.
In actual scenario, the field is checked whether it exists at all in the json. However, in the fake client uses the model attributes to verify, which are marked as None if they are not present. The client only checks for the attribute in the object and not the value for 'exists' condition and returns. This is not consistent with the actual scenario and returns invalid results.

#### Issue Type
- Bug Report

#### Steps to reproduce

* Create a Repository object with an attribute missing and add it to the Fake controller/client
* Search the repository with the condition that above missing attribute exists.

#### Actual results

Repository with the missing attribute is returned in the result.

#### Expected results

Only the repositories with the attribute having some value should be returned

rajulkumar:
match_field_exists returns true if an attribute of the
model object is none and that attribute is checked if
it exists. This change modifies to return true only
if the field is present and has some value.

fix #39

### DIFF
--- a/pubtools/pulplib/_impl/fake/match.py
+++ b/pubtools/pulplib/_impl/fake/match.py
@@ -112,7 +112,7 @@ def match_field_regex(matcher, field, obj):
 @visit(ExistsMatcher)
 def match_field_exists(_matcher, field, obj):
     value = get_field(field, obj)
-    return value is not ABSENT
+    return value not in [ABSENT, None]
 
 
 @visit(InMatcher)
--- a/tests/fake/test_fake_search.py
+++ b/tests/fake/test_fake_search.py
@@ -112,6 +112,28 @@ def test_search_created_exists():
     assert sorted(found) == [repo2, repo3]
 
 
+def test_search_with_model_field():
+    """search repos using model field 'created' in criteria.
+    returns only those repos which has the field and value is
+    not None
+    """
+    controller = FakeController()
+
+    repo1 = Repository(id="repo1")
+    repo2 = Repository(id="repo2", created=datetime.datetime.utcnow())
+    repo3 = Repository(id="repo3", created=None)
+
+    controller.insert_repository(repo1)
+    controller.insert_repository(repo2)
+    controller.insert_repository(repo3)
+
+    client = controller.client
+    crit = Criteria.with_field("created", Matcher.exists())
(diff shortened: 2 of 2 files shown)


pubtools/pulplib/_impl/fake/match.py
```diff
@@ -112,7 +112,7 @@ def match_field_regex(matcher, field, obj):
 @visit(ExistsMatcher)
 def match_field_exists(_matcher, field, obj):
     value = get_field(field, obj)
-    return value is not ABSENT
+    return value not in [ABSENT, None]
```
rohanpm: Not sure if it's going to be this easy.  Actually, see the TODO comment which exists later in this file about whether we need to differentiate between absent and present but None.

The issue here is that, if the value was None, you have no way of knowing if it was explicitly set to None or if it simply wasn't present.  For example, consider eng_product_id.  On a real Pulp server, it comes from field notes.eng_product.  That field could have been absent altogether, or it could have been present but with a value of null.  If it was present and null, then mongo $exists fragment would have found it.  So one can argue that exists() with the fake client should then *also* find objects where the field is None.

PulpObject classes simply don't hold the necessary info right now to differentiate between "this field is absent" and "this field is present but None", so I am not sure what's the most correct way to handle this right now.

If this change is solving some problem, it may make sense to proceed with it, as both old and new behaviors can be considered wrong from certain perspectives, so whichever works better in practice could be the better choice.

Either way, could we please get an autotest?

rohanpm: I thought about this some more and became more convinced that this behavior change you want to make is correct.  With fake client, our choices seem to be one of these caveats:

1. either it's not possible to represent an absent field
2. or it's not possible to represent a present-but-null field

I implemented it originally as (1) and you're now changing it to (2).  I guess that is actually the better choice simply because absent fields are very common while present-but-null fields are uncommon.

rajulkumar: Agree. And this is a problem only with exists matcher which not be very prevalent.
I'll update this patch with test cases

PR #41, opened 2019-08-21
Implement set maintenance API for pulp repositories

JayZ12138:
Also fix the bug of can't calculate checksum if file object contains
unicode during upload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Introduced ``Repository.is_temporary`` attribute
+- Support set maintenance mode of pulp repositories
+
+### Fixed
+- Fix can't calculate checksum if file object contains unicode data during upload
 
 ## [1.2.1] - 2019-08-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- Support set maintenance mode of pulp repositories
+- Support querying and updating maintenance mode of Pulp repositories
 
 ### Fixed
-- Fix can't calculate checksum if file object contains unicode data during upload
+- Fixed a crash in `upload_file` when passed a file object opened in text mode
 
 ## [1.3.0] - 2019-08-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-n/a
+### Added
+- Support set maintenance mode of pulp repositories
+
+### Fixed
+- Fix can't calculate checksum if file object contains unicode data during upload
 
 ## [1.3.0] - 2019-08-15
 
 ### Added
 
 - Introduced ``Repository.is_temporary`` attribute
-- Support set maintenance mode of pulp repositories
 - Extended search functionality; it is now possible to search using fields defined
   on the `PulpObject` classes. Searching on raw Pulp fields remains supported.
 
@@ -22,7 +25,6 @@ n/a
 - Fixed inconsistency between real and fake clients: both clients now immediately raise
   if a search is attempted with invalid criteria.  Previously, the fake client would
   instead return a failed `Future`.
-- Fix can't calculate checksum if file object contains unicode data during upload
 
 ## [1.2.1] - 2019-08-12
 
--- a/docs/api-reference.rst
+++ b/docs/api-reference.rst
@@ -7,6 +7,6 @@ API Reference
 
    api/client
    api/searching
+   api/maintenance
    api/model
    api/testing
-   api/maintenance
--- a/docs/api-reference.rst
+++ b/docs/api-reference.rst
@@ -9,3 +9,4 @@ API Reference
    api/searching
    api/model
    api/testing
+   api/maintenance
--- a/docs/api/maintenance.rst
+++ b/docs/api/maintenance.rst
@@ -1,4 +1,4 @@
-Maintenance Object
+Maintenance Mode
 ==================
 
 .. contents::
--- a/docs/api/maintenance.rst
+++ b/docs/api/maintenance.rst
@@ -4,8 +4,8 @@ Maintenance Mode
 .. contents::
   :local:
 
-.. autoclass:: pubtools.pulplib.MaintenanceEntry
+.. autoclass:: pubtools.pulplib.MaintenanceReport
     :members:
 
-.. autoclass:: pubtools.pulplib.MaintenanceReport
+.. autoclass:: pubtools.pulplib.MaintenanceEntry
     :members:
--- a/docs/api/maintenance.rst
+++ b/docs/api/maintenance.rst
@@ -1,0 +1,11 @@
+Maintenance Object
+==================
+
+.. contents::
+  :local:
+
+.. autoclass:: pubtools.pulplib.MaintenanceEntry
+    :members:
+
+.. autoclass:: pubtools.pulplib.MaintenanceReport
+    :members:
--- a/examples/set-maintenance
+++ b/examples/set-maintenance
@@ -1,21 +1,41 @@
 #!/usr/bin/env python
 import os
+import re
 import logging
 from argparse import ArgumentParser
 
-from pubtools.pulplib import Client
+from pubtools.pulplib import Client, Criteria, Matcher
 
 log = logging.getLogger("set-maintenance")
 
 
 def set_maintenance(client, regex):
 
-    report = client.set_maintainenance(regex)
-    log.info("maintenance mode set: %s", report)
+    report = client.get_maintenance_report().result()
 
-    client.set_maintainenance(regex, enable=False)
-    report = client.get_maintenance_report().json()
-    log.info("maintenance mode unset: %s", report)
+    crit = Criteria.with_field("id", Matcher.regex(regex))
+    # search repo
+    repos = self.search_repository(crit).result()
+
+    repo_ids = [repo.id for repo in repos.as_iter()]
+    report = report.add(repo_ids)
+    # get the repo ids and feed to report
+
+    tasks = client.set_maintainenance(report).result()
+
+    report = client.get_maintenance_report().result()
+    log.info("maintenance mode set: %s", report._json())
+
+    to_remove = []
+    for entry in report.entries:
+        if re.match(regex, entry.id):
+            to_remove.append(entry.id)
+
+    report = report.remove(to_remove)
+    tasks = client.set_maintainenance(report).result()
+
+    report = client.get_maintenance_report().result()
+    log.info("maintenance mode unset: %s", report._json())
 
 
 def make_client(args):
--- a/examples/set-maintenance
+++ b/examples/set-maintenance
@@ -1,0 +1,66 @@
+#!/usr/bin/env python
+import os
+import logging
+from argparse import ArgumentParser
+
+from pubtools.pulplib import Client
+
+log = logging.getLogger("set-maintenance")
+
+
+def set_maintenance(client, regex):
+
+    report = client.set_maintainenance(regex)
+    log.info("maintenance mode set: %s", report)
+
+    client.set_maintainenance(regex, enable=False)
+    report = client.get_maintenance_report().json()
+    log.info("maintenance mode unset: %s", report)
+
+
+def make_client(args):
+    auth = None
+
+    if args.username:
+        password = args.password
+        if password is None:
+            password = os.environ.get("PULP_PASSWORD")
+        if not password:
+            log.warning("No password provided for %s", args.username)
+        auth = (args.username, args.password)
+
+    return Client(args.url, auth=auth, verify=not args.insecure)
+
+
+def main():
+    log.setLevel(logging.INFO)
+    logging.basicConfig(format="%(message)s", level=logging.INFO)
+
+    parser = ArgumentParser(description="Set/unset maintenance mode of Pulp")
+    parser.add_argument("--url", help="Pulp server URL")
+    parser.add_argument("--username", help="Pulp username")
+    parser.add_argument(
+        "--password", help="Pulp password (or set PULP_PASSWORD in env)"
+    )
+    parser.add_argument("--debug", action="store_true")
+    parser.add_argument("--insecure", default=False, action="store_true")
+    parser.add_argument("--regex", action="store")
+
+    p = parser.parse_args()
+
+    if not p.url:
+        parser.error("--url is required")
+
+    if not p.regex:
+        parser.error("--regex is required")
+
+    if p.debug:
+        logging.getLogger("pubtools.pulplib").setLevel(logging.DEBUG)
+        log.setLevel(logging.DEBUG)
+
+    client = make_client(p)
+    return set_maintenance(client, p.regex)
+
+
+if __name__ == "__main__":
+    main()
--- a/examples/set-maintenance
+++ b/examples/set-maintenance
@@ -24,7 +24,7 @@ def set_maintenance(client, regex):
     tasks = client.set_maintainenance(report).result()
 
     report = client.get_maintenance_report().result()
-    log.info("maintenance mode set: %s", report._json())
+    log.info("maintenance mode set: %s", report.entries)
 
     to_remove = []
     for entry in report.entries:
@@ -35,7 +35,7 @@ def set_maintenance(client, regex):
     tasks = client.set_maintainenance(report).result()
 
     report = client.get_maintenance_report().result()
-    log.info("maintenance mode unset: %s", report._json())
+    log.info("maintenance mode unset: %s", report.entries)
 
 
 def make_client(args):
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -189,7 +189,7 @@ class Client(object):
         """Get the current maintenance mode status for this Pulp server.
 
         Returns:
-            Future[:class:`~pubtools.pulplib.MaintenceReport`]
+            Future[:class:`~pubtools.pulplib.MaintenanceReport`]
                 A future describes the maintenance status
         """
         report_ft = self._do_get_maintenance()
@@ -202,7 +202,7 @@ class Client(object):
         )
 
     def set_maintenance(self, report):
-        """Set the current maintenance mode status for this Pulp server.
+        """Set maintenance mode for this Pulp server.
 
         Args:
             report:
@@ -286,7 +286,7 @@ class Client(object):
             parsed = pulp_response.json()
         except Exception:
             # Couldn't parse as JSON?
-            # In other cases, if the response was unsuccessful, raise that.
+            # If the response was unsuccessful, raise that.
             # Otherwise re-raise parse error.
             pulp_response.raise_for_status()
             raise
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -4,13 +4,16 @@ import threading
 import hashlib
 from functools import partial
 
+import six
+from six.moves import StringIO
+
 import requests
 from more_executors import Executors
 from more_executors.futures import f_map, f_flat_map, f_return
 
 from ..page import Page
-from ..criteria import Criteria
-from ..model import Repository
+from ..criteria import Criteria, Matcher
+from ..model import Repository, MaintenanceReport
 from .search import filters_for_criteria
 from .errors import PulpException
 from .poller import TaskPoller
@@ -181,10 +184,87 @@ class Client(object):
             response_f, lambda data: self._handle_page(Repository, search, data)
         )
 
+    def get_maintenance_report(self):
+        """Get maintenance report from maintenance repository
+
+        Returns:
+            :class:`~pubtools.pulplib.MaintenceReport` object describes the
+            maintenance status
+
+            None if there's no report in the repository
+        """
+        try:
+            data = self._do_get_maintenance().result()
+        except requests.exceptions.HTTPError:
+            # if the maintenance mode hasn't been used before, then it could return
+            # HTTPError 404.
+            LOG.warning("No maintenance info found")
+            return None
+
+        return MaintenanceReport.from_data(data)
+
+    def set_maintainenance(self, regex, enable=True, **kwargs):
+        """Set repositories to maintenance mode.
+
+        Args:
+            regex (str):
+                A python regular expression pattern used to match repositories.
+
+            enable (bool):
+                If enable, then set the matched repositories to maintenance mode.
+                If not enable, unset the maintenance more for the matched repositories.
+
+            Optional keyword args:
+                owner (str):
+                    Person/party who set/unset maintenance mode.
+                message (str):
+                    Reason why set repositories to maintenance mode, doesn't have
+                    any effect if enable is False.
+
+        Return:
+            A result report in json format (str)
+        """
+        report = self.get_maintenance_report()
+        if not report:
+            if enable:
+                # create an empty report for later use.
+                report = MaintenanceReport()
+            else:
+                # if no maintenance info's found and user wants to unset
+                # maintenance mode, simply return None.
+                LOG.warning("No repository is in maintenance mode, exit")
+                return None
+
+        if enable:
+            # search repos match the regex and set to maintenance
+            crit = Criteria.with_field("id", Matcher.regex(regex))
+            repos = self.search_repository(crit).result()
+            repo_ids = [repo.id for repo in repos.as_iter()]
+            report.add(repo_ids, **kwargs)
+        elif not enable:
+            report.remove(regex, **kwargs)
+
+        repo = self.get_repository("redhat-maintenance").result()
+
+        # upload updated report to repository and publish
+        report_json = report.json()
+        report_fileobj = StringIO(report_json)
+        repo.upload_file(report_fileobj, "repos.json").result()
+        LOG.debug("Maintenance file uploaded in redhat-maintenance repository")
+
+        repo.publish().result()
+        LOG.debug("Maintenance status published")
+
+        # return the json dumped report, so caller could see the result directly.
+        return report_json
+
     def _do_upload_file(self, upload_id, file_obj, name):
         def do_next_upload(checksum, size):
             data = file_obj.read(self._CHUNK_SIZE)
             if data:
+                if isinstance(data, six.text_type):
+                    # if it's unicode, need to encode before calculate checksum
+                    data = data.encode("utf-8")
                 checksum.update(data)
                 return f_flat_map(
                     self._do_upload(data, upload_id, size),
@@ -363,3 +443,8 @@ class Client(object):
 
         LOG.debug("Deleting upload request %s", upload_id)
         return self._request_executor.submit(self._do_request, method="DELETE", url=url)
+
+    def _do_get_maintenance(self):
+        url = os.path.join(self._url, "pulp/isos/redhat-maintenance/repos.json")
+
+        return self._request_executor.submit(self._do_request, method="GET", url=url)
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -67,6 +67,9 @@ class Client(object):
     # overridden there.
     _RETRY_POLICY = retry.PulpRetryPolicy()
 
+    # the relative url where the maintenance report is stored
+    MAINTENANCE_REPORT_URL = "pulp/isos/redhat-maintenance/repos.json"
+
     def __init__(self, url, **kwargs):
         """
         Args:
@@ -185,7 +188,7 @@ class Client(object):
         )
 
     def get_maintenance_report(self):
-        """Get maintenance report from maintenance repository
+        """Get the current maintenance mode status for this Pulp server.
 
         Returns:
             Future[:class:`~pubtools.pulplib.MaintenceReport`]
@@ -195,21 +198,21 @@ class Client(object):
 
         return f_map(
             report_ft,
-            lambda data: MaintenanceReport.from_data(data)
+            lambda data: MaintenanceReport._from_data(data)
             if data
             else MaintenanceReport(),
         )
 
     def set_maintenance(self, report):
-        """Set repositories to maintenance mode.
+        """Set the current maintenance mode status for this Pulp server.
 
         Args:
             report:
                 An updated :class:`~pubtools.pulplib.MaintenanceReport` object that
                 will be used as the newest maintenance report.
         Return:
-            Future[:class:`~pubtools.pulplib.Task`]
-                A future which is resolved when publish succeeds.
+            Future[list[:class:`~pubtools.pulplib.Task`]]
+                A future which is resolved when maintenance mode has been updated successfully.
 
                 The future contains a task triggered and awaited during the publish
                 maintenance repository operation.
@@ -284,18 +287,17 @@ class Client(object):
             parsed = pulp_response.json()
         except Exception:
             # Couldn't parse as JSON?
-            if (
-                pulp_response.url.endswith(retry.MAINTENANCE_REPROT_URL)
-                and pulp_response.status_code == 404
+            if pulp_response.status_code == 404 and str(pulp_response.url).endswith(
+                cls.MAINTENANCE_REPORT_URL
             ):
-                # It's possible it's querying the maintenance report and it doesn't
-                # exist, then it means no repository is in maintenance.
+                # for getting maintenance report, it's possible the report
+                # doesn't exist, means no repositories are in maintenance mode,
+                # return None
                 return None
-            else:
-                # In other cases, if the response was unsuccessful, raise that.
-                # Otherwise re-raise parse error.
-                pulp_response.raise_for_status()
-                raise
+            # In other cases, if the response was unsuccessful, raise that.
+            # Otherwise re-raise parse error.
+            pulp_response.raise_for_status()
+            raise
 
         if (
             isinstance(parsed, dict)
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -2,8 +2,9 @@ import os
 import logging
 import threading
 import hashlib
+import json
 from functools import partial
-
+from concurrent.futures import Future
 import six
 from six.moves import StringIO
 
@@ -66,9 +67,6 @@ class Client(object):
     # This is mainly provided here as a hook for autotests, so the policy can be
     # overridden there.
     _RETRY_POLICY = retry.PulpRetryPolicy()
-
-    # the relative url where the maintenance report is stored
-    MAINTENANCE_REPORT_URL = "pulp/isos/redhat-maintenance/repos.json"
 
     def __init__(self, url, **kwargs):
         """
@@ -217,7 +215,8 @@ class Client(object):
                 The future contains a task triggered and awaited during the publish
                 maintenance repository operation.
         """
-        report_fileobj = StringIO(report._json())
+        report_json = json.dumps(report._export_dict(), indent=4, sort_keys=True)
+        report_fileobj = StringIO(report_json)
 
         repo = self.get_repository("redhat-maintenance").result()
 
@@ -287,13 +286,6 @@ class Client(object):
             parsed = pulp_response.json()
         except Exception:
             # Couldn't parse as JSON?
-            if pulp_response.status_code == 404 and str(pulp_response.url).endswith(
-                cls.MAINTENANCE_REPORT_URL
-            ):
-                # for getting maintenance report, it's possible the report
-                # doesn't exist, means no repositories are in maintenance mode,
-                # return None
-                return None
             # In other cases, if the response was unsuccessful, raise that.
             # Otherwise re-raise parse error.
             pulp_response.raise_for_status()
@@ -420,6 +412,33 @@ class Client(object):
         return self._request_executor.submit(self._do_request, method="DELETE", url=url)
 
     def _do_get_maintenance(self):
+        def map_404_to_none(response):
+            """Given a future, if it resolved with a 404 error then return a Future[None],
+            else return original future"""
+            out = Future()
+
+            def handle_response(ft):
+                exception = ft.exception()
+                if (
+                    hasattr(exception, "response")
+                    and exception.response.status_code == 404
+                ):
+                    # failed with 404 => returned future should return None
+                    out.set_result(None)
+                elif exception:
+                    # failed with other exception => returned future should raise that exception
+                    out.set_exception(exception)
+                else:
+                    # succeeded => returned future should pass through the result
+                    out.set_result(ft.result())
+
+            response.add_done_callback(handle_response)
+            return out
+
         url = os.path.join(self._url, "pulp/isos/redhat-maintenance/repos.json")
 
-        return self._request_executor.submit(self._do_request, method="GET", url=url)
+        response = self._request_executor.submit(
+            self._do_request, method="GET", url=url
+        )
+
+        return map_404_to_none(response)
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -12,7 +12,7 @@ from more_executors import Executors
 from more_executors.futures import f_map, f_flat_map, f_return
 
 from ..page import Page
-from ..criteria import Criteria, Matcher
+from ..criteria import Criteria
 from ..model import Repository, MaintenanceReport
 from .search import filters_for_criteria
 from .errors import PulpException
@@ -188,75 +188,40 @@ class Client(object):
         """Get maintenance report from maintenance repository
 
         Returns:
-            :class:`~pubtools.pulplib.MaintenceReport` object describes the
-            maintenance status
-
-            None if there's no report in the repository
+            Future[:class:`~pubtools.pulplib.MaintenceReport`]
+                A future describes the maintenance status
         """
-        try:
-            data = self._do_get_maintenance().result()
-        except requests.exceptions.HTTPError:
-            # if the maintenance mode hasn't been used before, then it could return
-            # HTTPError 404.
-            LOG.warning("No maintenance info found")
-            return None
+        report_ft = self._do_get_maintenance()
 
-        return MaintenanceReport.from_data(data)
+        return f_map(
+            report_ft,
+            lambda data: MaintenanceReport.from_data(data)
+            if data
+            else MaintenanceReport(),
+        )
 
-    def set_maintainenance(self, regex, enable=True, **kwargs):
+    def set_maintenance(self, report):
         """Set repositories to maintenance mode.
 
         Args:
-            regex (str):
-                A python regular expression pattern used to match repositories.
-
-            enable (bool):
-                If enable, then set the matched repositories to maintenance mode.
-                If not enable, unset the maintenance more for the matched repositories.
-
-            Optional keyword args:
-                owner (str):
-                    Person/party who set/unset maintenance mode.
-                message (str):
-                    Reason why set repositories to maintenance mode, doesn't have
-                    any effect if enable is False.
-
+            report:
+                An updated :class:`~pubtools.pulplib.MaintenanceReport` object that
+                will be used as the newest maintenance report.
         Return:
-            A result report in json format (str)
-        """
-        report = self.get_maintenance_report()
-        if not report:
-            if enable:
-                # create an empty report for later use.
-                report = MaintenanceReport()
-            else:
-                # if no maintenance info's found and user wants to unset
-                # maintenance mode, simply return None.
-                LOG.warning("No repository is in maintenance mode, exit")
-                return None
+            Future[:class:`~pubtools.pulplib.Task`]
+                A future which is resolved when publish succeeds.
 
-        if enable:
-            # search repos match the regex and set to maintenance
-            crit = Criteria.with_field("id", Matcher.regex(regex))
-            repos = self.search_repository(crit).result()
-            repo_ids = [repo.id for repo in repos.as_iter()]
-            report.add(repo_ids, **kwargs)
-        elif not enable:
-            report.remove(regex, **kwargs)
+                The future contains a task triggered and awaited during the publish
+                maintenance repository operation.
+        """
+        report_fileobj = StringIO(report._json())
 
         repo = self.get_repository("redhat-maintenance").result()
 
         # upload updated report to repository and publish
-        report_json = report.json()
-        report_fileobj = StringIO(report_json)
-        repo.upload_file(report_fileobj, "repos.json").result()
-        LOG.debug("Maintenance file uploaded in redhat-maintenance repository")
+        upload_ft = repo.upload_file(report_fileobj, "repos.json")
 
-        repo.publish().result()
-        LOG.debug("Maintenance status published")
-
-        # return the json dumped report, so caller could see the result directly.
-        return report_json
+        return f_flat_map(upload_ft, lambda _: repo.publish())
 
     def _do_upload_file(self, upload_id, file_obj, name):
         def do_next_upload(checksum, size):
@@ -319,10 +284,18 @@ class Client(object):
             parsed = pulp_response.json()
         except Exception:
             # Couldn't parse as JSON?
-            # If the response was unsuccessful, raise that.
-            # Otherwise re-raise parse error.
-            pulp_response.raise_for_status()
-            raise
+            if (
+                pulp_response.url.endswith(retry.MAINTENANCE_REPROT_URL)
+                and pulp_response.status_code == 404
+            ):
+                # It's possible it's querying the maintenance report and it doesn't
+                # exist, then it means no repository is in maintenance.
+                return None
+            else:
+                # In other cases, if the response was unsuccessful, raise that.
+                # Otherwise re-raise parse error.
+                pulp_response.raise_for_status()
+                raise
 
         if (
             isinstance(parsed, dict)
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -191,6 +191,8 @@ class Client(object):
         Returns:
             Future[:class:`~pubtools.pulplib.MaintenanceReport`]
                 A future describes the maintenance status
+
+        .. versionadded:: 1.4.0
         """
         report_ft = self._do_get_maintenance()
 
@@ -214,6 +216,8 @@ class Client(object):
 
                 The future contains a task triggered and awaited during the publish
                 maintenance repository operation.
+
+        .. versionadded:: 1.4.0
         """
         report_json = json.dumps(report._export_dict(), indent=4, sort_keys=True)
         report_fileobj = StringIO(report_json)
--- a/pubtools/pulplib/_impl/client/retry.py
+++ b/pubtools/pulplib/_impl/client/retry.py
@@ -11,6 +11,8 @@ LOG = logging.getLogger("pubtools.pulplib")
 ATTEMPTS = int(os.environ.get("PUBTOOLS_PULPLIB_RETRY_ATTEMPTS", "10"))
 SLEEP = float(os.environ.get("PUBTOOLS_PULPLIB_RETRY_SLEEP", "1.0"))
 MAX_SLEEP = float(os.environ.get("PUBTOOLS_PULPLIB_RETRY_MAX_SLEEP", "120.0"))
+# the relative url where the maintenance report is stored
+MAINTENANCE_REPROT_URL = "pulp/isos/redhat-maintenance/repos.json"
 
 
 class PulpRetryPolicy(RetryPolicy):
@@ -27,7 +29,16 @@ class PulpRetryPolicy(RetryPolicy):
         # at the error in more detail in order to decide whether to retry.
         retry = self._delegate.should_retry(attempt, future)
 
-        if future.exception() and retry:
+        exception = future.exception()
+        if exception and hasattr(exception, "response"):
+            # if request is querying maintenance repo, it's possible there's
+            # no report existed, not wasting time on retry.
+            url = str(exception.response.url)
+            status_code = exception.response.status_code
+            if url.endswith(MAINTENANCE_REPROT_URL) and status_code == 404:
+                return False
+
+        if exception and retry:
             self._log_retry(attempt, future)
 
         return retry
--- a/pubtools/pulplib/_impl/client/retry.py
+++ b/pubtools/pulplib/_impl/client/retry.py
@@ -11,8 +11,6 @@ LOG = logging.getLogger("pubtools.pulplib")
 ATTEMPTS = int(os.environ.get("PUBTOOLS_PULPLIB_RETRY_ATTEMPTS", "10"))
 SLEEP = float(os.environ.get("PUBTOOLS_PULPLIB_RETRY_SLEEP", "1.0"))
 MAX_SLEEP = float(os.environ.get("PUBTOOLS_PULPLIB_RETRY_MAX_SLEEP", "120.0"))
-# the relative url where the maintenance report is stored
-MAINTENANCE_REPROT_URL = "pulp/isos/redhat-maintenance/repos.json"
 
 
 class PulpRetryPolicy(RetryPolicy):
@@ -31,11 +29,8 @@ class PulpRetryPolicy(RetryPolicy):
 
         exception = future.exception()
         if exception and hasattr(exception, "response"):
-            # if request is querying maintenance repo, it's possible there's
-            # no report existed, not wasting time on retry.
-            url = str(exception.response.url)
-            status_code = exception.response.status_code
-            if url.endswith(MAINTENANCE_REPROT_URL) and status_code == 404:
+            # if returned status code is 404, never retry on that
+            if exception.response.status_code == 404:
                 return False
 
         if exception and retry:
--- a/pubtools/pulplib/_impl/model/maintenance.py
+++ b/pubtools/pulplib/_impl/model/maintenance.py
@@ -1,6 +1,5 @@
 import logging
 import datetime
-import json
 import os
 
 import jsonschema
@@ -25,7 +24,7 @@ def iso_time_now():
 class MaintenanceEntry(object):
     """Details about the maintenance status of a specific repository."""
 
-    repository_id = attr.ib(type=str)
+    repo_id = attr.ib(type=str)
     """ID of repository in maintenance.
     Note: there is no guarantee that a repository of this ID currenly exists
           in the Pulp server."""
@@ -65,6 +64,14 @@ class MaintenanceReport(object):
     If empty, then it means no repositories are in maintenance mode.
     """
 
+    @entries.validator
+    def check_duplicates(self, attribute, value):
+        # pylint: disable=unused-argument
+        # check if there's duplicate entries
+        repo_ids = [entry.repo_id for entry in value]
+        if len(repo_ids) != len(set(repo_ids)):
+            raise ValueError("Duplicate entries")
+
     @classmethod
     def _from_data(cls, data):
         """Create a new report with raw data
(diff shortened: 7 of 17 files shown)


pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,147 @@
...
+    _SCHEMA = load_schema("maintenance")
+
+    last_updated = attr.ib(default=_iso_time_now(), type=str)
+    """:class:`~datetime.datetime` in UTC when this report was last updated,
+    if it's the first time the report is created, current time is used."""
+
+    last_updated_by = attr.ib(default="Content Delivery", type=str)
+    """Person/party who updated the report last time"""
+
+    repos = attr.ib(default={}, type=dict)
```
rohanpm: I know that these are stored as a dict on the underlying data format, but I'd suggest not doing that here.  Mainly because it gets you this weird asymmetric data structure, as mentioned in an earlier comment, where each element in the dict carries all info about the maintenance mode entry *except* the repo ID itself, so you have to pass that around separately.

I suggest a MaintenanceEntry class for each item in this container, and that class should have a repo ID, and this should be a tuple of entries rather than a dict.  Why tuple: because those are immutable.

```suggestion
    entries = attr.ib(default=attr.Factory(tuple), type=tuple)
    """A list of maintenance mode entries.  Each entry contains metadata for a single repository
    currently in maintenance mode."""
```

To keep the invariant that there are no duplicate entries for a single repo, a validator could be implemented.

Also incidentally, not 100% sure about this, but I guess setting default={} with attrs is unsafe for the same reason as setting some default={} as a function/method argument is unsafe - presumably all the instances get a reference to the *same* dict, which could be modified and would then affect many objects.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,147 @@
...
+    """A dictionary contains repo_id: :class:`MaintainedRepo` pairs, indicates
+    which repositories are in maintenance mode and details.
+    If empty, then it means no repositories is in maintenance mode.
+    """
+
+    @classmethod
+    def from_data(cls, data):
+        """Create a new report with a dictionary"""
```
rohanpm: Can you please mention & link to the schema here?  (You should include it in docs)

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,147 @@
...
+            last_updated=data["last_updated"],
+            last_updated_by=data["last_updated_by"],
+            repos={},
+        )
+        for repo_id, details in data["repos"].items():
+            maintenance.repos[repo_id] = MaintainedRepo(**details)
+
+        return maintenance
+
+    def json(self):
```
rohanpm: I suggest not making this public API (or do you have a specific case in mind where something needs this?)

Of course, you need to be able to convert it to a raw dict locally, but we should discourage any other code except for this library making use of that raw format.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,147 @@
...
+                        "message": details.message,
+                        "owner": details.owner,
+                        "started": details.started,
+                    }
+                }
+            )
+
+        return json.dumps(report, indent=4, sort_keys=True)
+
+    def add(self, repo_ids, **kwargs):
```
rohanpm: Per the suggestion to make this class immutable - I think this should rather be creating and returning a modified copy of this object instead of modifying self directly.  This helps keep code thread-safe.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,147 @@
...
+
+            Optional keyword args:
+                message (str):
+                    Reason why put the repo to maintenance.
+                owner (str):
+                    Who set the maintenance mode.
+
+        """
+        message = kwargs.get("message") or "Maintenance mode is enabled"
+        owner = kwargs.get("owner") or "Content Delivery"
```
rohanpm: I suggest "pubtools-pulplib" is a better name than "Content Delivery".  It more clearly communicates that we only know which tool did the update and we don't know anything else.

However, maybe we can do even better again.  What do you think of instead defaulting to something like $USER@$HOSTNAME?  e.g. "root@rcm-pub-worker-01.example.com" - would give us a better hint at what set a repo in maintenance mode.

In practice, of course we want the callers to provide this argument whenever possible.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,147 @@
...
+            )
+
+        self._update_owner_time(owner)
+
+    def remove(self, regex, **kwargs):
+        """Remove entries from the maintenece report. Once the entry is removed,
+        it means the corresponding repository isn't in maintenance mode anymore.
+
+        Args:
+            regex (str):
```
rohanpm: Why regex again?  It has the same problem I mentioned before that it's actually quite awkward to use, if you know exactly which repos you want to remove.

pubtools/pulplib/_impl/schema/maintenance.yaml
```diff
@@ -0,0 +1,53 @@
+title: maintenance report
```
rohanpm: There should be an update to schema .rst file to include this in docs as well.

CHANGELOG.md
```diff
@@ -7,12 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-n/a
+### Added
+- Support set maintenance mode of pulp repositories
```
rohanpm:
```suggestion
- Support querying and updating maintenance mode of Pulp repositories
```

CHANGELOG.md
```diff
@@ -7,12 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Fixed
+- Fix can't calculate checksum if file object contains unicode data during upload
```
rohanpm:
```suggestion
- Fixed a crash in `upload_file` when passed a file object opened in text mode
```

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +184,87 @@ def search_repository(self, criteria=None):
...
+
+            None if there's no report in the repository
+        """
+        try:
+            data = self._do_get_maintenance().result()
+        except requests.exceptions.HTTPError:
+            # if the maintenance mode hasn't been used before, then it could return
+            # HTTPError 404.
+            LOG.warning("No maintenance info found")
+            return None
```
rohanpm: I wouldn't recommend returning None, isn't it more user-friendly to return a maintenance report which shows that no repos are in maintenance mode?

If someone is writing code dealing with maintenance mode, is there any advantage in forcing them to write separate code for the None case vs "nothing is in maintenance mode" case?

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +184,87 @@ def search_repository(self, criteria=None):
...
+        elif not enable:
+            report.remove(regex, **kwargs)
+
+        repo = self.get_repository("redhat-maintenance").result()
+
+        # upload updated report to repository and publish
+        report_json = report.json()
+        report_fileobj = StringIO(report_json)
+        repo.upload_file(report_fileobj, "repos.json").result()
+        LOG.debug("Maintenance file uploaded in redhat-maintenance repository")
```
rohanpm: For these kinds of messages, logging task IDs when available would also be nice, so we can connect up a Pulp task ID with a specific operation in the logs.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +184,87 @@ def search_repository(self, criteria=None):
...
+            data = self._do_get_maintenance().result()
+        except requests.exceptions.HTTPError:
+            # if the maintenance mode hasn't been used before, then it could return
+            # HTTPError 404.
+            LOG.warning("No maintenance info found")
+            return None
+
+        return MaintenanceReport.from_data(data)
+
+    def set_maintainenance(self, regex, enable=True, **kwargs):
```
rohanpm: A question, why do we have this public method for setting maintenance mode given some repo IDs, *and* MaintenanceReport also has public methods for adjusting the report given repo IDs?  Which one do we expect clients to use?

I think the cleaner design would be to have responsibilities split up so that:

- methods on MaintenanceReport are responsible for creating a new & modified report
- this method is only responsible for uploading a MaintenanceReport

It also gives you a nice symmetry - the client has a getter which returns a MaintenanceReport, then it also has a setter which returns a MaintenanceReport too.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +184,87 @@ def search_repository(self, criteria=None):
...
+        """
+        report = self.get_maintenance_report()
+        if not report:
+            if enable:
+                # create an empty report for later use.
+                report = MaintenanceReport()
+            else:
+                # if no maintenance info's found and user wants to unset
+                # maintenance mode, simply return None.
+                LOG.warning("No repository is in maintenance mode, exit")
```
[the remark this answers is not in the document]
JayZ12138: Hm, I make it a warning because I think something could be wrong if someone wants to set to maintenance when there's even no maintenance report, but most probably, this could branch might won't be touched after maintenance is set up, so either debug or info is OK

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +184,87 @@ def search_repository(self, criteria=None):
...
+        # upload updated report to repository and publish
+        report_json = report.json()
+        report_fileobj = StringIO(report_json)
+        repo.upload_file(report_fileobj, "repos.json").result()
+        LOG.debug("Maintenance file uploaded in redhat-maintenance repository")
+
+        repo.publish().result()
+        LOG.debug("Maintenance status published")
+
+        # return the json dumped report, so caller could see the result directly.
```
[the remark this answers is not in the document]
JayZ12138: it doesn't have to return, I was just thinking it would be easier for the caller to see the maintenance status directly, instead of checking the content in pulp.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,147 @@
...
+
+LOG = logging.getLogger("pubtools.pulplib")
+
+
+def _iso_time_now():
+    return datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+@attr.s(kw_only=True)
+class MaintainedRepo(object):
```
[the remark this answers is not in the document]
JayZ12138: yeah, I feel the same. I found it's difficult naming them reasonably.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +184,87 @@ def search_repository(self, criteria=None):
...
+        Returns:
+            :class:`~pubtools.pulplib.MaintenceReport` object describes the
+            maintenance status
+
+            None if there's no report in the repository
+        """
+        try:
+            data = self._do_get_maintenance().result()
+        except requests.exceptions.HTTPError:
+            # if the maintenance mode hasn't been used before, then it could return
```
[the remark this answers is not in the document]
JayZ12138: Could you elaborate it about how to use submit_retry? should I modify the PulpRetryPolicy directly?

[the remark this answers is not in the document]
rohanpm: Hmm... with submit_retry,  you would create a new custom retry policy and then replace the submit call here with a call to submit_retry.  But, thinking about it more, it seems you can forget about submit_retry and you can just modify PulpRetryPolicy.should_retry directly.

I don't think we've ever observed a case where there was a "flaky" 404 from Pulp which went away after retries, so it should be fine to modify that so it never retries on 404.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +184,87 @@ def search_repository(self, criteria=None):
...
+                owner (str):
+                    Person/party who set/unset maintenance mode.
+                message (str):
+                    Reason why set repositories to maintenance mode, doesn't have
+                    any effect if enable is False.
+
+        Return:
+            A result report in json format (str)
+        """
+        report = self.get_maintenance_report()
```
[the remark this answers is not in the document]
JayZ12138: Hm, I don't get the point of making it non blocking. 

For upload or publish, I can get it: user could run many upload and publish tasks in parallel to improve the efficiency. But for set maintenance, with the current mechanism, it's not possible to run several maintenance tasks in parallel -- image if  two tasks download the maintenance file at the same time and update it with different content and upload, the result is the one uploaded first would be overwriten by the latter one. So, if we have to run set maintenance task in serial, why do we need to make it non blocking?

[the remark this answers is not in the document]
rohanpm: OK, there's actually two aspects to this and I was mixing them together wrongly, sorry about that.

Firstly, does it need to be non-blocking: no, strictly speaking it doesn't, for the reason you said.

Secondly, does it need to return a future: yes, I think it does, for the sake of consistency.  Right now, every method which creates Pulp tasks or issues requests to Pulp will return a Future.  If we start making exceptions to that based on implementation details, then the API becomes confusing very fast.  The caller needs to carefully remember the return type of every method and it will seem arbitrary ("getting a repo returns a Future... but getting maintenance report doesn't?")

So I will be happy as long as this returns a Future, it doesn't *have* to be achieved by writing it in non-blocking style.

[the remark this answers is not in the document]
JayZ12138: So, for set maintenance, what should be wrapped in the future returned? publish tasks or a result report object?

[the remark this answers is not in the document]
JayZ12138: for set maintenance, I plan to do it in blocking-style, since the non-blocking style code I can think of is ugly and hard to read:
`report_ft = self.get_maintenance_report()

        def process_report(report):
            # process report in threads
            if enable:
                # search repositories matched regex
                crit = Criteria.with_field("id", Matcher.regex(regex))
                repos_ft = self.search_repository(crit)
                repo_ids_ft = f_map(repos_ft, lambda repos: [repo.id for repo in repos.as_iter()])
                return f_map(repo_ids_ft, lambda repo_ids: report.add(repo_ids, **kwargs))
            else:
                return f_return(report.remove(regex, **kwargs))

        process_report_ft = f_flat_map(report_ft, process_report)

        repo_ft = self.get_repository("redhat-maintenance")

        # upload updated report to repository and publish
        report_fileobj = f_map(process_report_ft, lambda report: StringIO(report.json()))

        upload_ft = f_flat_map(
            f_sequence(repo_ft, process_report_ft),
            lambda repo_report: repo_report[0].upload_file(repo_report[1])
        )

        publish_ft = f_flat_map(
            f_sequence(repo_ft, upload_ft),
            lambda repo_upload: repo_upload[0].publish()
        )

        return publish_ft`

pubtools/pulplib/_impl/model/maintenance.py
JayZ12138: Hm, I was just thinking about providing an alternative format of data to the user, if it's discouraged, then I guess this should be made private.

pubtools/pulplib/_impl/model/maintenance.py
JayZ12138: Hm, set maintenance isn't likely to be run in threads, do we still need this?

pubtools/pulplib/_impl/model/maintenance.py
JayZ12138: for the second point about default to {} is unsafe because it might refer to the same object, I tested a little and seems like it's safe to use:
`>>> import attr
>>> @attr.s
... class c(object):
...     c = attr.ib(default={}, type=dict)
... 
>>> c
<class '__main__.c'>
>>> c()
c(c={})
>>> c(c={'a':'b'})
c(c={'a': 'b'})
>>> c()
c(c={})
`

pubtools/pulplib/_impl/client/client.py
JayZ12138: hm, I guess I got what you mean\

pubtools/pulplib/_impl/model/maintenance.py
rohanpm: Yeah, I think so.  Thread safety is only the most obvious reason for immutability, but there are others.

I expect in practice that it'll lead to better designed code in callers, with less side effects.  Code with side effects is generally more at risk of becoming unmaintainable, and also less likely to be idempotent (and retryable).

In any case, for this library the design choice of using immutable objects for classes representing Pulp data has already been made. We should stick with it throughout, unless there are very good reasons for an exception.

(If it's unclear whether to count this class as "Pulp data" since it's something we made up ourselves and not really a supported feature of Pulp - I don't think this should make a difference from the library user's point of view.  The API seems like it's successfully hiding implementation details here such that the user doesn't have to know or care whether maintenance mode is something built into Pulp, or emulated by our library.  That's a good thing.)

pubtools/pulplib/_impl/client/client.py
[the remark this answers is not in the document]
rohanpm: I don't have a strong opinion what should be in the returned future (just not the raw JSON).  A list of tasks created during the process seems the most consistent with other methods right now.

The main use case there would be simply logging which tasks were involved in updating maintenance mode, so you can cross reference with Pulp logs and understand the purpose of each task.

pubtools/pulplib/_impl/model/maintenance.py
rohanpm: The unsafe thing would be when you modify the contents of that dict.  I did my own testing just now and indeed it seems to be unsafe in that every instance shares the same dict, probably not intended:

```
>>> import attr
>>> @attr.s
... class Klass:
...   c = attr.ib(default={})
... 
>>> k1 = Klass()
>>> k2 = Klass()
>>> k1.c
{}
>>> k2.c
{}

# Now let's try assigning something to the dict

>>> k1.c['foo'] = 'bar'
>>> k1.c
{'foo': 'bar'}

# Oops, my change on k1 is also visible via k2
>>> k2.c
{'foo': 'bar'}
```

docs/api/maintenance.rst
```diff
@@ -0,0 +1,11 @@
+Maintenance Object
```
rohanpm:
```suggestion
Maintenance Mode
```

Probably makes more sense from user's point of view.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +184,52 @@ def search_repository(self, criteria=None):
             response_f, lambda data: self._handle_page(Repository, search, data)
         )
 
+    def get_maintenance_report(self):
+        """Get maintenance report from maintenance repository
```
rohanpm: Suggest a slight tweak to not expose implementation details:

```suggestion
        """Get the current maintenance mode status for this Pulp server.
```

i.e. no need to mention "maintenance repository".

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +184,52 @@ def search_repository(self, criteria=None):
...
+
+        return f_map(
+            report_ft,
+            lambda data: MaintenanceReport.from_data(data)
+            if data
+            else MaintenanceReport(),
+        )
+
+    def set_maintenance(self, report):
+        """Set repositories to maintenance mode.
```
rohanpm:
```suggestion
        """Set the current maintenance mode status for this Pulp server.
```

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +184,52 @@ def search_repository(self, criteria=None):
...
+
+        Args:
+            report:
+                An updated :class:`~pubtools.pulplib.MaintenanceReport` object that
+                will be used as the newest maintenance report.
+        Return:
+            Future[:class:`~pubtools.pulplib.Task`]
```
rohanpm: I think this is not accurate, is it?  Doesn't publish() return a list of tasks?

```suggestion
            Future[list[:class:`~pubtools.pulplib.Task`]]
```

Note: not 100% how to make it render in docs the best way, might need to play around.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +184,52 @@ def search_repository(self, criteria=None):
...
+                A future which is resolved when publish succeeds.
```
rohanpm: Again I would rather be a bit more vague and generic with respect to what's actually happening here.  No need to commit to a particular implementation here - we might need to add or remove steps to maintenance mode update in the future.

So, I think we shouldn't mention here about "maintenance repository" or "publish", e.g. :

```suggestion
                A future which is resolved when maintenance mode has been updated successfully.
                
                The future returns a list of zero or more Pulp tasks created during the update.
```

I see the main purpose of returning anything is just to allow for log messages matching up with Pulp tasks, such as:

```
Updated maintenance mode: <task1>, <task2>
```

...so that it's possible to look at Pulp task logs and find the matching pubtools-pulp log, and vice-versa.  The caller ideally won't depend on returned tasks being specifically publish tasks.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -239,10 +284,18 @@ def _unpack_response(cls, pulp_response):
             parsed = pulp_response.json()
         except Exception:
             # Couldn't parse as JSON?
-            # If the response was unsuccessful, raise that.
-            # Otherwise re-raise parse error.
-            pulp_response.raise_for_status()
-            raise
+            if (
+                pulp_response.url.endswith(retry.MAINTENANCE_REPROT_URL)
```
rohanpm:
```suggestion
                pulp_response.url.endswith(retry.MAINTENANCE_REPORT_URL)
```

The solution here is a bit ugly though, mixing a condition specific to maintenance mode in with this code applied on all responses.

Did you also try the suggestion of *never* retrying on 404?  Then I would expect do_get_maintenance should catch the error and return None, rather than adding this special casing here.

pubtools/pulplib/_impl/client/retry.py
```diff
@@ -11,6 +11,8 @@
 ATTEMPTS = int(os.environ.get("PUBTOOLS_PULPLIB_RETRY_ATTEMPTS", "10"))
 SLEEP = float(os.environ.get("PUBTOOLS_PULPLIB_RETRY_SLEEP", "1.0"))
 MAX_SLEEP = float(os.environ.get("PUBTOOLS_PULPLIB_RETRY_MAX_SLEEP", "120.0"))
+# the relative url where the maintenance report is stored
+MAINTENANCE_REPROT_URL = "pulp/isos/redhat-maintenance/repos.json"
```
rohanpm:
```suggestion
MAINTENANCE_REPORT_URL = "pulp/isos/redhat-maintenance/repos.json"
```

pubtools/pulplib/_impl/client/retry.py
```diff
@@ -27,7 +29,16 @@ def should_retry(self, attempt, future):
...
         retry = self._delegate.should_retry(attempt, future)
 
-        if future.exception() and retry:
+        exception = future.exception()
+        if exception and hasattr(exception, "response"):
+            # if request is querying maintenance repo, it's possible there's
+            # no report existed, not wasting time on retry.
+            url = str(exception.response.url)
+            status_code = exception.response.status_code
+            if url.endswith(MAINTENANCE_REPROT_URL) and status_code == 404:
```
rohanpm:
```suggestion
            if url.endswith(MAINTENANCE_REPORT_URL) and status_code == 404:
```

Similar as the comment elsewhere, do we actually need to connect this with maintenance mode?  Could we just never retry on 404 (at least, until we find some reason it needs a retry)?

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,147 @@
+import logging
```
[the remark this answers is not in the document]
rohanpm: Resolving this since you added a new page.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+OWNER = "%s@%s" % (USER, HOSTNAME)
+
+
+def iso_time_now():
+    return datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+@attr.s(kw_only=True, frozen=True)
+class MaintenanceEntry(object):
+    """Details about the maintenance"""
```
rohanpm:
```suggestion
    """Details about the maintenance status of a specific repository."""
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+
+    id = attr.ib(default=None, type=str)
```
rohanpm: Could we please call this "repository_id"?  I think bare "id" is risky - if maintenance entries are later linked to any other kind of entity which has an "id" - like an outage ID - then it becomes ambiguous.

Also, should it really have default=None?  I think this is probably the one attribute which is absolutely mandatory for every entry, and an entry without the repo ID would be unusable.

```suggestion
    repository_id = attr.ib(type=str)
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+    """Why this repository is in maintenance"""
+    owner = attr.ib(default=None, type=str)
+    """Who set this repository in maintenance mode"""
+    started = attr.ib(default=None, type=datetime.datetime)
+    """:class:`~datetime.datetime` in UTC at when the maintenance started"""
+
+
+@attr.s(kw_only=True, frozen=True)
+class MaintenanceReport(object):
+    """Represents the maintenance status of Pulp repositories
```
rohanpm: Current doc style is to use full sentences with punctuation, so:

```suggestion
    """Represents the maintenance status of Pulp repositories.
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+
+    _SCHEMA = load_schema("maintenance")
+
+    last_updated = attr.ib(default=iso_time_now(), type=datetime.datetime)
+    """:class:`~datetime.datetime` in UTC when this report was last updated,
+    if it's the first time the report is created, current time is used."""
+
+    last_updated_by = attr.ib(default=None, type=str)
+    """Person/party who updated the report last time"""
```
rohanpm:
```suggestion
    """Person/party who updated the report last time."""
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+
+    entries = attr.ib(default=attr.Factory(tuple), type=tuple)
+    """A tuple contains :class:`MaintenanceEntry` objects, indicates
```
rohanpm:
```suggestion
    """A tuple of :class:`MaintenanceEntry` objects, indicating
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+    which repositories are in maintenance mode and details.
+    If empty, then it means no repositories is in maintenance mode.
```
rohanpm:
```suggestion
    If empty, then it means no repositories are in maintenance mode.
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+
+    @classmethod
+    def from_data(cls, data):
```
rohanpm: I would rather not make this one public API (or do you have a particular reason to do so?)

For the other PulpObject which do have a from_data, the main reason was because there's a lot of existing code for querying those objects already which aren't using this library, and it's valuable to be able to connect up with that code (e.g. use Repository class with existing code querying Pulp via pa_tool).

But for maintenance mode, I don't see much reason to provide a way for any other code to parse the raw data at all.  Allowing others to do that will essentially lock us in to the specific format of repos.json, as we couldn't change it without breaking this API.  I would rather avoid that, and try to have all of the code understanding that format built into this library only, as much as possible.

So, could we make this private too unless you have specific use-case for it?

(Note if I'm wrong about this, we can always promote it to public easily later.)

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+                A list of repository ids. New entries with these repository ids will
+                be added to the maintenance report.
+
+            Optional keyword args:
+                message (str):
+                    Reason why put the repo to maintenance.
+                owner (str):
+                    Who set the maintenance mode.
+
+        """
```
rohanpm:
```suggestion
        Returns:
           :class:`~pubtools.pulplib.MaintenanceReport`
              A copy of this maintenance report with added repositories.
        """
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+from ..schema import load_schema
+from .common import InvalidDataException
+
+
+LOG = logging.getLogger("pubtools.pulplib")
+
+
+USER = os.environ.get("USER")
+HOSTNAME = os.environ.get("HOSTNAME")
+OWNER = "%s@%s" % (USER, HOSTNAME)
```
rohanpm: Environment variables aren't guaranteed, so can you please have a default in case they're missing?

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+    _SCHEMA = load_schema("maintenance")
+
+    last_updated = attr.ib(default=iso_time_now(), type=datetime.datetime)
+    """:class:`~datetime.datetime` in UTC when this report was last updated,
+    if it's the first time the report is created, current time is used."""
+
+    last_updated_by = attr.ib(default=None, type=str)
+    """Person/party who updated the report last time"""
+
+    entries = attr.ib(default=attr.Factory(tuple), type=tuple)
```
rohanpm: I'm not seeing any protection against duplicate entries for a repo.  Could you please add that somewhere?

Most likely should be done via an attrs validator.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+def iso_time_now():
+    return datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+@attr.s(kw_only=True, frozen=True)
+class MaintenanceEntry(object):
+    """Details about the maintenance"""
+
+    id = attr.ib(default=None, type=str)
+    """ID of repository in maintenance"""
```
rohanpm: In our current solution, there's actually no guarantee that the repo IDs in the maintenance report match existing repos.  So, this could point at a repository which doesn't exist.

I'm OK with that caveat of the design, but I think it's worthwhile mentioning in our docs somewhere.  Maybe right here is the place?  Something like:

```suggestion
    """ID of repository in maintenance.
    Note: there is no guarantee that a repository of this ID currenly exists in the Pulp server."""
```

pubtools/pulplib/_impl/model/maintenance.py
rohanpm: Resolving this since the requested change was made.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+                    Reason why put the repo to maintenance.
+                owner (str):
+                    Who set the maintenance mode.
+
+        """
+        message = kwargs.get("message") or "Maintenance mode is enabled"
+        owner = kwargs.get("owner") or OWNER
+
+        to_add = []
+        for repo in repo_ids:
```
rohanpm: I think we need some kind of handling here in the case the caller asked to add a repo which is already in the report.  Right now, seems like that'd create duplicates (which can't actually be stored).  I guess it would make more sense to overwrite the earlier entry (e.g. updating the comment for a repo which is already in maintenance mode).

Slightly wacky implementation idea - what if you did this:

```
out = MaintenanceReport(last_updated_by=owner, entries=tuple(entries))
return MaintenanceReport._from_json(out._json())
```

i.e. you do a round-trip through the serialization format before you return the object:

- means you can't have bugs where you return a MaintenanceReport which can't actually be serialized or which "changes" during serialization (e.g. duplicate entries)
- based on implementation of `_json`, this will already de-duplicate entries without any extra code needed

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+                    id=repo, owner=owner, message=message, started=iso_time_now()
+                )
+            )
+        entries = list(self.entries)
+        entries.extend(to_add)
+
+        return MaintenanceReport(last_updated_by=owner, entries=tuple(entries))
+
+    def remove(self, repo_ids, **kwargs):
+        """Remove entries from the maintenece report. Once the entry is removed,
```
rohanpm:
```suggestion
        """Remove entries from the maintenance report. Once the entry is removed,
```

Also, I think that text about "once the entry is removed" could be misleading... it seems like someone could read that with understanding that modifying this object immediately modifies the data in Pulp, not realizing they need to do an extra call for that.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+
+        Args:
+            repo_ids (list[str]):
+                A list of repository ids. Entries match repository ids will be removed
+                from the maintenance report.
+
+            Optional keyword args:
+                owner (str):
+                    Who unset the maintenance mode.
+        """
```
rohanpm:
```suggestion
        Returns:
           :class:`~pubtools.pulplib.MaintenanceReport`
              A copy of this maintenance report with removed repositories.
        """
```
```

File: pubtools/pulplib/_impl/model/maintenance.py [hunk 41]
```diff
@@ -0,0 +1,179 @@
...
+        owner = kwargs.get("owner") or OWNER
+
+        repo_ids = set(repo_ids)
+        # convert to set, make checking faster
+        to_remove = []
```
rohanpm: I think you could do this in a single loop and (in my opinion) would be cleaner:

```suggestion
        new_entries = []
        for entry in self.entries:
          if entry.id not in repo_ids:
            new_entries.append(entry)

        return attr.evolve(self, entries=tuple(new_entries))
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,179 @@
...
+        for repo in repo_ids:
+            to_add.append(
+                MaintenanceEntry(
+                    id=repo, owner=owner, message=message, started=iso_time_now()
+                )
+            )
+        entries = list(self.entries)
+        entries.extend(to_add)
+
+        return MaintenanceReport(last_updated_by=owner, entries=tuple(entries))
```
rohanpm: Really minor, but it's slightly more maintainable if you use attr.evolve instead which is designed to return a new copy with selected fields changed:

```suggestion
        return attr.evolve(self, entries=tuple(entries))
```

The benefit is that if MaintenanceReport gets a new field later, it will automatically be covered here.  Without that, there's a risk someone forgets to update this code, and so add/remove methods here have an unintentional side-effect of resetting some fields back to default.

tests/client/test_client.py
```diff
@@ -165,3 +176,60 @@ def test_get_missing(client, requests_mocker):
...
+    requests_mocker.post(
+        "https://pulp.example.com/pulp/api/v2/repositories/search/",
+        [{"json": [{"id": "redhat-maintenance", "notes": {"_repo-type": "iso-repo"}}]}],
+    )
+
+    report = MaintenanceReport()
+
+    with patch("pubtools.pulplib.FileRepository.upload_file") as mocked_upload:
+        with patch("pubtools.pulplib.Repository.publish") as mocked_publish:
+            mocked_upload.return_value = f_return()
```
rohanpm: Test isn't really thorough enough, the uploaded content is never tested.  Seems like the test would pass even if the code uploaded garbage every time?

tests/fake/test_fake_set_maintenance.py
```diff
@@ -0,0 +1,76 @@
...
+
+from pubtools.pulplib import (
+    FakeController,
+    Repository,
+    FileRepository,
+    MaintenanceReport,
+    InvalidDataException,
+)
+
+OWNER = "%s@%s" % (os.environ.get("USER"), os.environ.get("HOSTNAME"))
```
rohanpm: I think you should be overriding these during the test, rather than taking them from the environment (where they might be missing altogether).

You can use monkeypatch.setenv for setting an environment variable for the duration of a test.  It would however require you to adjust the code slightly to read the environment variables more than once (not only at compile time).

tests/repository/test_upload.py
```diff
@@ -130,3 +131,24 @@ def test_get_relative_url_with_file_object(tmpdir):
         repo._get_relative_url(file_obj, "some/path/")
 
     assert repo._get_relative_url(file_obj, "path/foo.txt") == "path/foo.txt"
+
+
+def test_upload_file_contains_unicode(client, requests_mocker):
+    file_obj = StringIO(u"some unicode")
```
rohanpm: Ideally you should put some real non-ASCII text in here, because u"some unicode" can be interpreted without problems by code which can only handle ASCII or latin1 - so this test doesn't really ensure that unicode data is handled correctly.

Test would be a little stronger if you threw in an emoticon, some Chinese or whatever.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -239,10 +284,18 @@ def _unpack_response(cls, pulp_response):
             parsed = pulp_response.json()
         except Exception:
```
rohanpm: I was looking at the Travis CI failure, which is in pylint (broad-except rule) - but you didn't introduce this code so it was unclear why it just fails now.

Then I realized the reason:  with the previous code, the exception would always be unconditionally re-raised (if a different exception wasn't raised earlier).  pylint is smart enough to figure that out.

Now the code is changed in a way which is too dynamic for pylint to consider it acceptable (from pylint's point of view, *any* kind of exception could be caught and thrown away here).  You'll need to find a way to avoid this error.

I think ideally, this method shouldn't have any logic specific to maintenance mode.

examples/set-maintenance
```diff
@@ -0,0 +1,86 @@
...
+    repos = self.search_repository(crit).result()
+
+    repo_ids = [repo.id for repo in repos.as_iter()]
+    report = report.add(repo_ids)
+    # get the repo ids and feed to report
+
+    tasks = client.set_maintainenance(report).result()
+
+    report = client.get_maintenance_report().result()
+    log.info("maintenance mode set: %s", report._json())
```
rohanpm: Please avoid using the private API in the example.

Isn't the plain string representation giving you all the same data anyway, in arguably more readable form?

examples/set-maintenance
```diff
@@ -0,0 +1,86 @@
...
+    to_remove = []
+    for entry in report.entries:
+        if re.match(regex, entry.id):
+            to_remove.append(entry.id)
+
+    report = report.remove(to_remove)
+    tasks = client.set_maintainenance(report).result()
+
+    report = client.get_maintenance_report().result()
+    log.info("maintenance mode unset: %s", report._json())
```
rohanpm: Avoid usage of private API.

pubtools/pulplib/_impl/model/maintenance.py
rohanpm: No, that's not the only possible way to get duplicates.  This class is public API and users of this library can create instances of it themselves with whatever entries they want.  In particular when working with the fake client, I think they would be expected to do that (i.e. the use-case is just set the maintenance mode on the fake Pulp server to some specific canned data before running a certain test).

So, I think the validator is still valuable to prevent anyone writing code which creates invalid objects.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,193 @@
...
+
+        report = attr.evolve(
+            self,
+            entries=tuple(entries),
+            last_updated_by=owner,
+            last_updated=iso_time_now(),
+        )
+
+        # filter out duplicate entries by export json
+        return MaintenanceReport._from_data(json.loads(report._json()))
```
rohanpm: OK, I should have noticed this earlier, this is showing an odd design mismatch now:

Why is it that, when deserializing from JSON to a MaintenanceReport, from_data accepts a raw dict and so the job of decoding JSON into a dict lives elsewhere... but when *serializing*, MaintenanceReport doesn't produce a raw dict but instead produces a JSON string?  It seems like an odd inconsistency.  Would it be better if it both accepted and produced raw dicts, instead of accepting a raw dict but producing a JSON string?

This isn't really important though, given that both are now private API.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,193 @@
...
+
+def iso_time_now():
+    return datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+@attr.s(kw_only=True, frozen=True)
+class MaintenanceEntry(object):
+    """Details about the maintenance status of a specific repository."""
+
+    repository_id = attr.ib(type=str)
```
rohanpm: Sorry, I made a mistake here.  In Task class, the same kind of attribute was called "repo_id".  So, even though it goes against the guideline of avoiding abbreviations, consistency is more important and we should in fact call this:

```suggestion
    repo_id = attr.ib(type=str)
```

pubtools/pulplib/_impl/client/client.py
```diff
@@ -239,7 +287,14 @@ def _unpack_response(cls, pulp_response):
...
             # Couldn't parse as JSON?
-            # If the response was unsuccessful, raise that.
+            if pulp_response.status_code == 404 and str(pulp_response.url).endswith(
+                cls.MAINTENANCE_REPORT_URL
```
rohanpm: I still very much would like to see a solution to this which doesn't involve putting maintenance mode specific logic into this generic unpack_response method.  It seems like _do_get_maintenance itself should be catching the 404 and translating it into None.

Doing it here is messy and risks problems later.  For example, you don't really know that the executor will never be used to fetch some *other* URL which happens to end with "pulp/isos/redhat-maintenance/repos.json", for a reason other than maintenance mode.  Imagine a script trying to do something like "download all files from a FileRepository", and then the error handling behavior for this one file is different for no good reason.

I think what you would need to do is something like this, in _do_get_maintenance:

```
response = self._request_executor.submit(self._do_request, method="GET", url=url)
# If we get a 404, translate it to None instead
return _map_404_to_none(response)

...

def _map_404_to_none(response):
  """Given a future, if it resolved with a 404 error then return a Future[None], else return original future"""
  out = Future()

  def handle_response(f):
    exception = f.exception()
    if hasattr(exception, 'response') and response.status_code == 404:
      # failed with 404 => returned future should return None
      out.set_result(None)
    elif exception:
      # failed with other exception => returned future should raise that exception
      out.set_exception(exception)
    else:
      # succeeded => returned future should pass through the result
      out.set_result(f.result())

  response.add_done_callback(handle_response)
  return out
```

Ideally, more-executors would have had some utility function to make this easier, but currently it doesn't, map/flatmap are only designed for the successful case.

Actually... what I wrote above is to support non-blocking style.  But I think you chose to make it blocking anyway, so couldn't you write it even simpler by making _do_get_maintenance also blocking and then simply try/except for the exception you want to handle?

pubtools/pulplib/_impl/client/retry.py
```diff
@@ -27,7 +27,13 @@ def should_retry(self, attempt, future):
         # at the error in more detail in order to decide whether to retry.
         retry = self._delegate.should_retry(attempt, future)
 
-        if future.exception() and retry:
+        exception = future.exception()
+        if exception and hasattr(exception, "response"):
+            # if returned status code is 404, never retry on that
+            if exception.response.status_code == 404:
+                return False
```
rohanpm: Looks like this line is uncovered, better add a test.

pubtools/pulplib/_impl/client/client.py
rohanpm: I saw it in email and had trouble finding it on github.  If you would share the code for that attempt, I could try to figure out why it didn't work, but otherwise not sure.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -239,7 +286,7 @@ def _unpack_response(cls, pulp_response):
             parsed = pulp_response.json()
         except Exception:
             # Couldn't parse as JSON?
-            # If the response was unsuccessful, raise that.
+            # In other cases, if the response was unsuccessful, raise that.
```
rohanpm: This change doesn't make sense now, should put it back.
```suggestion
            # If the response was unsuccessful, raise that.
```

pubtools/pulplib/_impl/client/client.py
```diff
@@ -363,3 +410,35 @@ def _delete_upload_request(self, upload_id):
...
+                exception = ft.exception()
+                if (
+                    hasattr(exception, "response")
+                    and exception.response.status_code == 404
+                ):
+                    # failed with 404 => returned future should return None
+                    out.set_result(None)
+                elif exception:
+                    # failed with other exception => returned future should raise that exception
+                    out.set_exception(exception)
```
rohanpm: Now this seems to be uncovered, so there should be a case where fetching the JSON file fails with (e.g.) 500 error.

pubtools/pulplib/_impl/client/retry.py
rohanpm: From running locally, it seems to be covered now.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +185,53 @@ def search_repository(self, criteria=None):
...
+        """
+        report_json = json.dumps(report._export_dict(), indent=4, sort_keys=True)
+        report_fileobj = StringIO(report_json)
+
+        repo = self.get_repository("redhat-maintenance").result()
+
+        # upload updated report to repository and publish
+        upload_ft = repo.upload_file(report_fileobj, "repos.json")
+
+        return f_flat_map(upload_ft, lambda _: repo.publish())
```
rohanpm: Hmm... I don't think this is correct actually.  repo.publish() returns a *list of futures*, right?  Not a single future.  flat map isn't going to be able to flatten a list of futures into a single future.

So, what does the function actually return right now?  A Future[Task], Future[Future[Task]], Future[list[Future[Task]]] ... something else?

tests/client/test_client.py
```diff
@@ -165,3 +176,81 @@ def test_get_missing(client, requests_mocker):
...
+        [{"json": [{"id": "redhat-maintenance", "notes": {"_repo-type": "iso-repo"}}]}],
+    )
+
+    report = MaintenanceReport._from_data(maintenance_report)
+
+    with patch("pubtools.pulplib.FileRepository.upload_file") as mocked_upload:
+        with patch("pubtools.pulplib.Repository.publish") as mocked_publish:
+            mocked_upload.return_value = f_return()
+            mocked_publish.return_value = f_return()
+            client.set_maintenance(report).result()
```
rohanpm: Related to my question about what this function actually returns: you never tested the returned value here.  How do you know it doesn't return garbage?

Can you please either test the returned value, or make it return nothing meaningful (e.g. Future[None]).

pubtools/pulplib/_impl/client/client.py
rohanpm: Resolving as this now seems fixed.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,200 @@
...
+    """Person/party who updated the report last time."""
+
+    entries = attr.ib(default=attr.Factory(tuple), type=tuple)
+    """A tuple of :class:`MaintenanceEntry` objects, indicating
+    which repositories are in maintenance mode and details.
+    If empty, then it means no repositories are in maintenance mode.
+    """
+
+    @entries.validator
+    def check_duplicates(self, attribute, value):
```
rohanpm: This should be named with _ as well as it's not public API.

Also, for the pylint issue, personally I prefer the style of naming the argument as simply _.  Although it does have the downside that someone has to read the validator docs to understand what that unused arg is.

```suggestion
    def _check_duplicates(self, _, value):
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,200 @@
...
+                        "started": entry.started,
+                    }
+                }
+            )
+
+        return report
+
+    def add(self, repo_ids, **kwargs):
+        """Add entries to maintenance report and update the timestamp. Every
+        entry added to the report represents a repository in maintenace mode.
```
rohanpm:
```suggestion
        entry added to the report represents a repository in maintenance mode.
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,200 @@
...
+
+        Args:
+            repo_ids (list[str]):
+                A list of repository ids. New entries with these repository ids will
+                be added to the maintenance report.
+
+                Note: it's users' responsibility to make sure the repository exists in
+                the pulp server, this method doesn't check the repository's existence.
```
rohanpm: "repository's existence" is technically wrong grammar for referring to multiple repositories.  It's a bit awkward to get the correct wording for this in English.  Also, Pulp should use capital P.  Here's my attempt:

```suggestion
                the Pulp server, this method doesn't check for the existence of repositories.
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,200 @@
...
+    return datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+@attr.s(kw_only=True, frozen=True)
+class MaintenanceEntry(object):
+    """Details about the maintenance status of a specific repository."""
+
+    repo_id = attr.ib(type=str)
+    """ID of repository in maintenance.
+    Note: there is no guarantee that a repository of this ID currenly exists
```
rohanpm:
```suggestion
    Note: there is no guarantee that a repository of this ID currently exists
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,200 @@
...
+          in the Pulp server."""
```
rohanpm: Have you published the docs locally and reviewed them?  This doesn't render correctly at the moment due to current whitespace.  Can you please fix that?

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,200 @@
...
+    message = attr.ib(default=None, type=str)
+    """Why this repository is in maintenance."""
+    owner = attr.ib(default=None, type=str)
+    """Who set this repository in maintenance mode."""
+    started = attr.ib(default=iso_time_now(), type=datetime.datetime)
```
rohanpm: I was reviewing docs and noticed something: due to this default, the documentation for this class actually contains the time at which the docs were generated :)

At first I thought it's just a funny thing, but thinking about it some more - this shows that the default is a bad idea, I think.  The reason is that a piece of code creating a MaintenanceEntry will create different objects depending on when that code is run.

For example:

```
   entry1 = MaintenanceEntry(repo_id='repo1', owner='some-user')
   entry2 = MaintenanceEntry(repo_id='repo1', owner='some-user')

   # question: is this true or false?
   # answer: depends how much time passed between above two lines!
   assert entry1 == entry2
```

I would say it violates reasonable expectations for this kind of "plain old data" class - two identical pieces of code for creating an object, should give you two identical objects, and not have a hidden dependency on the clock.

So, could we please set the default here to None, and handle this some other way?  e.g. setting the started value during `add`, if needed.

(If there's a question "why would anyone manually create a MaintenanceEntry anyway" - I'd expect the main reason to do that is when writing autotests related to maintenance mode features.)

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,200 @@
...
+
+        Args:
+            repo_ids (list[str]):
+                A list of repository ids. New entries with these repository ids will
+                be added to the maintenance report.
+
+                Note: it's users' responsibility to make sure the repository exists in
+                the pulp server, this method doesn't check the repository's existence.
+
+            Optional keyword args:
```
rohanpm: I don't think this quite renders correctly in the docs.  When I look at the docs, I see:

* ``repo_ids (list[str]) –``
* ``keyword args (Optional) –``

I believe it's being interpreted as "an argument named 'keyword args', of type Optional".  The arguments documented under here aren't formatted consistently with other arguments either, e.g. str for message isn't a link.

I'd suggest actually the best way to handle this is to give these arguments the same headers as others - no "Optional keyword args" subsection - and just mention something like "(optional)" in the documentation for each one.

pubtools/pulplib/_impl/model/maintenance.py
rohanpm: Resolving since we have validator now.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +185,53 @@ def search_repository(self, criteria=None):
             response_f, lambda data: self._handle_page(Repository, search, data)
         )
 
+    def get_maintenance_report(self):
+        """Get the current maintenance mode status for this Pulp server.
+
+        Returns:
+            Future[:class:`~pubtools.pulplib.MaintenceReport`]
```
rohanpm: Class name is misspelled.

```suggestion
            Future[:class:`~pubtools.pulplib.MaintenanceReport`]
```

docs/api-reference.rst
```diff
@@ -9,3 +9,4 @@ API Reference
    api/searching
    api/model
    api/testing
+   api/maintenance
```
rohanpm: I read the docs and I think it flows oddly now, with maintenance being the last section.  Could we move it up to just under "searching" instead?  That way it flows like: (1) general introduction to client; (2) docs about a particular thing you can do with client (searching), (3) docs about another thing you can do with client (manipulate maintenance mode).

Testing could also be considered a "thing you can do with client", but it's kind of a special case.

docs/api/maintenance.rst
```diff
@@ -0,0 +1,11 @@
+Maintenance Mode
+==================
+
+.. contents::
+  :local:
+
+.. autoclass:: pubtools.pulplib.MaintenanceEntry
+    :members:
+
+.. autoclass:: pubtools.pulplib.MaintenanceReport
```
rohanpm: Could we please swap the order of the classes here?  MaintenanceReport doc string contains some general explanation of what maintenance mode is, so it seems to make more sense for that to come first.

pubtools/pulplib/_impl/client/client.py
rohanpm: Nope, I forgot how the API works.  repo.publish() returns a *future of list*.  So this should be OK.

tests/client/test_client.py
rohanpm: I figured out I was wrong about that other comment.  But the point still seems to be valid that the returned value is never checked here?

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +185,53 @@ def search_repository(self, criteria=None):
...
+
+        Args:
+            report:
+                An updated :class:`~pubtools.pulplib.MaintenanceReport` object that
+                will be used as the newest maintenance report.
+        Return:
+            Future[list[:class:`~pubtools.pulplib.Task`]]
+                A future which is resolved when maintenance mode has been updated successfully.
+
+                The future contains a task triggered and awaited during the publish
```
rohanpm:
```suggestion
                The futures contain tasks triggered and awaited during the publish
```

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -0,0 +1,206 @@
...
+    will be blocked.  Other operations remain possible.
+
+    This object holds information on the set of repositories currently in maintenance mode.
+    """
+
+    _OWNER = "%s@%s" % (USER, HOSTNAME) if all([USER, HOSTNAME]) else "pubtools.pulplib"
+
+    _SCHEMA = load_schema("maintenance")
+
+    last_updated = attr.ib(default=iso_time_now(), type=datetime.datetime)
```
rohanpm: I see, having default=now will have the same problem here as I mentioned for MaintenanceEntry class.  Two identical statements would produce different objects depending on timing.  Can we please have a None default here too?

```suggestion
    last_updated = attr.ib(default=None, type=datetime.datetime)
```

tests/client/test_client.py
rohanpm: I'm not saying that it needs a very detailed test.  But right now, you've documented it as returning a Future[list[Task]], and this test only verifies that it returns a Future[?].  So there is no protection at all against it returning something entirely bogus, like Future[None], Future[str] or whatever...

Actually I think it would be fair to simply test it like this:

```suggestion
            f = client.set_maintenance(report)
            f.result()
            # It should return *exactly* what publish returned
            assert f is mocked_publish.return_value
```

As you say, publish is tested elsewhere.  So if we implement this API as "it returns the result of publish", it seems fair to directly test that.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,10 +185,53 @@ def search_repository(self, criteria=None):
...
+    def set_maintenance(self, report):
+        """Set maintenance mode for this Pulp server.
+
+        Args:
+            report:
+                An updated :class:`~pubtools.pulplib.MaintenanceReport` object that
+                will be used as the newest maintenance report.
+        Return:
+            Future[list[:class:`~pubtools.pulplib.Task`]]
+                A future which is resolved when maintenance mode has been updated successfully.
```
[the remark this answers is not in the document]
rohanpm: Oops, I messed this up.  This suggestion isn't valid, current docs are correct.

JayZ12138:
> I think this is really close. Jay, do you want it to be merged & released as soon as it's approved?

Yeah. 
Guess for some new APIs, now I can safely say they are new in version 1.5.0?
How about the Maintenance-related class? How should I indicate they are new in 1.5.0?

tests/client/test_client.py
JayZ12138: Hm, it's not completely the same. 
The set_maintenance returns a FlatMapFuture object, which wraps upload and publish. mocked_publish returns a Future, so they can't be same..., though the results are identical. 

I tried to assert f == f_flat_map(mocked_upload(), lambda _: mocked_publish()), but it gave me this error:
>               assert f == f_flat_map(mocked_upload(), lambda _: mocked_publish())
E               AssertionError: assert <FlatMapFuture at 0x7f7f54c19b50 state=finished returned NoneType> == <FlatMapFuture at 0x7f7f54c19e10 state=finished returned NoneType>
E                +  where <FlatMapFuture at 0x7f7f54c19e10 state=finished returned NoneType> = f_flat_map(<Future at 0x7f7f54c19990 state=finished returned NoneType>, <function <lambda> at 0x7f7f54cadb90>)
E                +    where <Future at 0x7f7f54c19990 state=finished returned NoneType> = <MagicMock name='upload_file' id='140184859088656'>()

Can we check if the result of set_maintenance is whatever wrapped in f_return()?

JayZ12138:
> > I think this is really close. Jay, do you want it to be merged & released as soon as it's approved?
> 
> Yeah.
> Guess for some new APIs in client.py, now I can safely say they are new in version 1.5.0?
> How about the Maintenance-related class? How should I indicate they are new in 1.5.0?

Hm, looks like you're waiting for my patch, I'll do 1.4.0

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,6 +185,50 @@ def search_repository(self, criteria=None):
             response_f, lambda data: self._handle_page(Repository, search, data)
         )
 
+    def get_maintenance_report(self):
+        """Get the current maintenance mode status for this Pulp server.
+
+        Returns:
+            Future[:class:`~pubtools.pulplib.MaintenanceReport`]
+                A future describes the maintenance status
```
rohanpm:
```suggestion
                A future describing the maintenance status
```

pubtools/pulplib/_impl/client/client.py
```diff
@@ -181,6 +185,50 @@ def search_repository(self, criteria=None):
...
+
+                The future contains a task triggered and awaited during the publish
+                maintenance repository operation.
+
+        .. versionadded:: 1.4.0
+        """
+        report_json = json.dumps(report._export_dict(), indent=4, sort_keys=True)
+        report_fileobj = StringIO(report_json)
+
+        repo = self.get_repository("redhat-maintenance").result()
```
rohanpm: As an aside (not arguing to change it now), I'm wondering if maybe we should have just let this library also create the repo when it doesn't exist.  I'm not much benefit to the current design where we require some other tool to be run first.

rohanpm:
Last few messages here were due to me trying a couple different ways to resolve conflicts before merge.

PR #43, opened 2019-08-29
Added search_distributor api and Matcher.less_than matcher

rajulkumar:
- search_distributor api to get distributors on defined criteria
- less_than(value) matcher to get objects with field value less than 'value'

### DIFF
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -171,12 +171,12 @@ class Client(object):
         )
 
     def search_distributor(self, criteria=None):
-        """Search the distributors matching the given criteria
+        """Search the distributors matching the given criteria.
 
         Args:
             criteria (:class:`~pubtools.pulplib.Criteria`)
                 A criteria object used for this search.
-                If None, search for all repositories.
+                If None, search for all distributors.
 
         Returns:
             Future[:class:`~pubtools.pulplib.Page`]
@@ -194,8 +194,7 @@ class Client(object):
             "filters": filters_for_criteria(criteria, return_type),
         }
         search = {"criteria": pulp_crit}
-        if search_options:
-            search.update(search_options)
+        search.update(search_options or {})
 
         response_f = self._do_search(resource_type, search)
 
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -10,7 +10,7 @@ from more_executors.futures import f_map, f_flat_map, f_return
 
 from ..page import Page
 from ..criteria import Criteria
-from ..model import Repository
+from ..model import Repository, Distributor
 from .search import filters_for_criteria
 from .errors import PulpException
 from .poller import TaskPoller
@@ -165,20 +165,45 @@ class Client(object):
                 Each page will contain a collection of
                 :class:`~pubtools.pulplib.Repository` objects.
         """
+        search_options = {"distributors": True}
+        return self._search(
+            Repository, "repositories", criteria=criteria, search_options=search_options
+        )
+
+    def search_distributor(self, criteria=None):
+        """Search the distributors matching the given criteria
+
+        Args:
+            criteria (:class:`~pubtools.pulplib.Criteria`)
+                A criteria object used for this search.
+                If None, search for all repositories.
+
+        Returns:
+            Future[:class:`~pubtools.pulplib.Page`]
+                A future representing the first page of results.
+
+                Each page will contain a collection of
+                :class:`~pubtools.pulplib.Distributor` objects.
+        """
+        return self._search(Distributor, "distributors", criteria=criteria)
+
+    def _search(self, return_type, resource_type, criteria=None, search_options=None):
         pulp_crit = {
             "skip": 0,
             "limit": self._PAGE_SIZE,
-            "filters": filters_for_criteria(criteria, Repository),
+            "filters": filters_for_criteria(criteria, return_type),
         }
-        search = {"criteria": pulp_crit, "distributors": True}
+        search = {"criteria": pulp_crit}
+        if search_options:
+            search.update(search_options)
 
-        response_f = self._do_search("repositories", search)
+        response_f = self._do_search(resource_type, search)
 
         # When this request is resolved, we'll have the first page of data.
         # We'll need to convert that into a page and also keep going with
         # the search if there's more to be done.
         return f_map(
-            response_f, lambda data: self._handle_page(Repository, search, data)
+            response_f, lambda data: self._handle_page(return_type, search, data)
         )
 
     def _do_upload_file(self, upload_id, file_obj, name):
--- a/pubtools/pulplib/_impl/client/search.py
+++ b/pubtools/pulplib/_impl/client/search.py
@@ -7,10 +7,12 @@ from pubtools.pulplib._impl.criteria import (
     EqMatcher,
     InMatcher,
     ExistsMatcher,
+    LessThanMatcher
 )
 
 from pubtools.pulplib._impl import compat_attr as attr
 from pubtools.pulplib._impl.model.attr import PULP2_FIELD, PY_PULP2_CONVERTER
+from pubtools.pulplib._impl.util import _is_iso_date_format
 
 
 def all_subclasses(klass):
@@ -86,5 +88,13 @@ def field_match(to_match):
 
     if isinstance(to_match, ExistsMatcher):
         return {"$exists": True}
+
+    if isinstance(to_match, LessThanMatcher):
+        value = to_match._value
+        # only value of the format YYYY-mm-ddTHH:MM:SSZ is
+        # treated as date
+        if _is_iso_date_format(value):
+            return {"$lt": {"$date": value}}
+        return {"$lt": value}
 
     raise TypeError("Not a matcher: %s" % repr(to_match))
--- a/pubtools/pulplib/_impl/client/search.py
+++ b/pubtools/pulplib/_impl/client/search.py
@@ -29,6 +29,15 @@ def to_mongo_json(value):
     if isinstance(value, datetime.datetime):
         return {"$date": value.strftime("%Y-%m-%dT%H:%M:%SZ")}
 
+    if isinstance(value, (list, tuple)):
+        return [to_mongo_json(elem) for elem in value]
+
+    if isinstance(value, dict):
+        out = {}
+        for (key, val) in value.items():
+            out[key] = to_mongo_json(val)
+        return out
+
     return value
 
 
--- a/pubtools/pulplib/_impl/client/search.py
+++ b/pubtools/pulplib/_impl/client/search.py
@@ -34,7 +34,7 @@ def to_mongo_json(value):
 
 def map_field_for_type(field_name, matcher, type_hint):
     if not type_hint:
-        return (field_name, matcher)
+        return None
 
     attrs_classes = all_subclasses(type_hint)
     attrs_classes = [cls for cls in attrs_classes if attr.has(cls)]
@@ -55,7 +55,7 @@ def map_field_for_type(field_name, matcher, type_hint):
         raise NotImplementedError("Searching on field %s is not supported" % field_name)
 
     # No match => no change, search exactly what was requested
-    return (field_name, matcher)
+    return None
 
 
 def filters_for_criteria(criteria, type_hint=None):
@@ -78,7 +78,9 @@ def filters_for_criteria(criteria, type_hint=None):
         field = criteria._field
         matcher = criteria._matcher
 
-        field, matcher = map_field_for_type(field, matcher, type_hint)
+        mapped = map_field_for_type(field, matcher, type_hint)
+        if mapped:
+            field, matcher = mapped
 
         return {field: field_match(matcher)}
 
--- a/pubtools/pulplib/_impl/client/search.py
+++ b/pubtools/pulplib/_impl/client/search.py
@@ -7,7 +7,7 @@ from pubtools.pulplib._impl.criteria import (
     EqMatcher,
     InMatcher,
     ExistsMatcher,
-    LessThanMatcher
+    LessThanMatcher,
 )
 
 from pubtools.pulplib._impl import compat_attr as attr
--- a/pubtools/pulplib/_impl/client/search.py
+++ b/pubtools/pulplib/_impl/client/search.py
@@ -1,3 +1,4 @@
+import datetime
 from pubtools.pulplib._impl.criteria import (
     AndCriteria,
     OrCriteria,
@@ -12,7 +13,6 @@ from pubtools.pulplib._impl.criteria import (
 
 from pubtools.pulplib._impl import compat_attr as attr
 from pubtools.pulplib._impl.model.attr import PULP2_FIELD, PY_PULP2_CONVERTER
-from pubtools.pulplib._impl.util import _is_iso_date_format
 
 
 def all_subclasses(klass):
@@ -21,6 +21,15 @@ def all_subclasses(klass):
     for subclass in klass.__subclasses__():
         out.update(all_subclasses(subclass))
     return out
+
+
+def to_mongo_json(value):
+    # Return a value converted to the format expected for a mongo JSON
+    # expression. Only a handful of special types need explicit conversions.
+    if isinstance(value, datetime.datetime):
+        return {"$date": value.strftime("%Y-%m-%dT%H:%M:%SZ")}
+
+    return value
 
 
 def map_field_for_type(field_name, matcher, type_hint):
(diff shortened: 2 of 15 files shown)


pubtools/pulplib/_impl/client/client.py
```diff
@@ -165,20 +165,45 @@ def search_repository(self, criteria=None):
                 Each page will contain a collection of
                 :class:`~pubtools.pulplib.Repository` objects.
         """
+        search_options = {"distributors": True}
+        return self._search(
+            Repository, "repositories", criteria=criteria, search_options=search_options
+        )
+
+    def search_distributor(self, criteria=None):
+        """Search the distributors matching the given criteria
```
rohanpm:
```suggestion
        """Search the distributors matching the given criteria.
```

pubtools/pulplib/_impl/client/client.py
```diff
@@ -165,20 +165,45 @@ def search_repository(self, criteria=None):
...
+
+        Args:
+            criteria (:class:`~pubtools.pulplib.Criteria`)
+                A criteria object used for this search.
+                If None, search for all repositories.
```
rohanpm:
```suggestion
                If None, search for all distributors.
```

pubtools/pulplib/_impl/client/client.py
```diff
@@ -165,20 +165,45 @@ def search_repository(self, criteria=None):
...
         pulp_crit = {
             "skip": 0,
             "limit": self._PAGE_SIZE,
-            "filters": filters_for_criteria(criteria, Repository),
+            "filters": filters_for_criteria(criteria, return_type),
         }
-        search = {"criteria": pulp_crit, "distributors": True}
+        search = {"criteria": pulp_crit}
+        if search_options:
+            search.update(search_options)
```
rohanpm: A trivial suggestion: in cases like this you can consider instead dropping the "if" and writing

```suggestion
      search.update(search_options or {})
```

The difference being that you're not creating a new branch of code counted separately by coverage tools.  I would usually but not always count that as a benefit.

pubtools/pulplib/_impl/client/search.py
```diff
@@ -87,4 +89,12 @@ def field_match(to_match):
     if isinstance(to_match, ExistsMatcher):
         return {"$exists": True}
 
+    if isinstance(to_match, LessThanMatcher):
+        value = to_match._value
+        # only value of the format YYYY-mm-ddTHH:MM:SSZ is
+        # treated as date
+        if _is_iso_date_format(value):
```
rohanpm: Interesting.  I'm not sure this is right, because we have some fields stored in Pulp which are genuinely stored as strings in the ISO8601 format, and not stored as dates.  Seems like this might make them unsearchable with LessThan, as there's no way to opt-out of using $date operator here if the value looks like ISO8601 format.

I think notes.created is an example of such a field.

Question: can we just base this on the type of the value instead?  I mean if _value is a datetime.datetime, then it's appropriate to search with "$date" operator.  Otherwise it's not.  I don't like the library "guessing" that a passed string should be interpreted as a date.

pubtools/pulplib/_impl/criteria.py
```diff
@@ -219,6 +219,28 @@ def in_(cls, values):
         """
         return InMatcher(values)
 
+    @classmethod
+    def less_than(cls, value):
+        """
+        Returns a matcher for a field whose value is less than the specified input
+        value
```
rohanpm:
```suggestion
        value.
```

pubtools/pulplib/_impl/criteria.py
```diff
@@ -219,6 +219,28 @@ def in_(cls, values):
...
+
+        Arguments:
+            values (object)
```
rohanpm:
```suggestion
            value (object)
```

pubtools/pulplib/_impl/criteria.py
```diff
@@ -219,6 +219,28 @@ def in_(cls, values):
...
+                An object to match against the field
+
+        Example:
+            .. code-block:: python
+
+            # would match where last_publish is before "2019-08-27T00:00:00Z
```
rohanpm:
```suggestion
            # would match where last_publish is before "2019-08-27T00:00:00Z"
```

pubtools/pulplib/_impl/criteria.py
```diff
@@ -219,6 +219,28 @@ def in_(cls, values):
...
+            # date is reqired to be in this format
```
rohanpm:
```suggestion
            # date is required to be in this format
```

But anyway, I'm probably not agreeing with the example (commented below).

pubtools/pulplib/_impl/criteria.py
```diff
@@ -219,6 +219,28 @@ def in_(cls, values):
...
+            crit = Criteria.with_field(
+                'last_publish',
+                Matcher.less_than("2019-08-27T00:00:00Z")
```
rohanpm: I think if last_publish is stored as a datetime in Pulp/mongo, then we should also be passing a datetime.datetime here.  Can that be made to work?

pubtools/pulplib/_impl/fake/match.py
```diff
@@ -124,6 +126,21 @@ def match_in(matcher, field, obj):
     return False
 
 
+@visit(LessThanMatcher)
+def match_less(matcher, field, obj):
+    # for datetime.dateime fields, obj_value is returned as pulp_value(iso format)
+    # if the model filed name is same as pulp filed name else returns a obj_value
```
rohanpm:
```suggestion
    # if the model field name is same as pulp field name else returns a obj_value
```

pubtools/pulplib/_impl/fake/match.py
```diff
@@ -124,6 +126,21 @@ def match_in(matcher, field, obj):
...
+    # as datetime.datetime object
+    obj_value = get_field(field, obj)
```
rohanpm: I guess this part will need refactoring per my suggestion that caller should be able to pass a datetime.datetime, not a str, when searching datetimes.

pubtools/pulplib/_impl/schema/repository.yaml
```diff
@@ -17,6 +17,9 @@ definitions:
       distributor_type_id:
         # String ID of distributor's type, e.g. "yum_distributor"
         type: string
+      repo_id:
+        # String ID of the repository the distributor is attached
```
rohanpm:
```suggestion
        # String ID of distributor's repository
```

tests/client/test_search.py
```diff
@@ -58,6 +58,19 @@ def test_field_regex_criteria():
     ) == {"some.field": {"$regex": "abc"}}
 
 
+def test_field_less_than_criteria():
+    """with_filed with less_than is translated as expected for
```
rohanpm:
```suggestion
    """with_field with less_than is translated as expected for
```

tests/client/test_search.py
```diff
@@ -58,6 +58,19 @@ def test_field_regex_criteria():
...
+    date and non-date types
+    """
+    c1 = Criteria.with_field("num_field", Matcher.less_than("5"))
+    c2 = Criteria.with_field("date_field", Matcher.less_than("2019-08-27T00:00:00Z"))
```
rohanpm: This spot again would need changing to support passing a datetime.datetime.

I think this test shows the inconsistency of current setup - when caller passes less_than("5"), the client doesn't guess that 5 should be interpreted as an int.  But when caller passes less_than("2019-08-27T00:00:00Z"), it does guess that it should be interpreted as a date.

pubtools/pulplib/_impl/client/search.py
rohanpm: I did some playing around with this locally and found:

- it's pretty easy to make it accept datetime.datetime objects from the caller. 
 Quick and dirty example: https://gist.github.com/rohanpm/88931ee34b3c0c4cefed0bd933766af2
- it's also possible to support searching by *both* datetime and str if we like.  It's not included in above gist.  It can be done by taking the above change, and then adding a bit more to support converting a str into a datetime, around the bit of the code where PY_PULP2_CONVERTER is currently accessed.

rajulkumar: I think using datetime.datetime object as input for the matcher will surely cut down the guessing for date. Wasn't sure if that's an option.

Agree that matcher._value type is used as the source of truth for the field type to get the desired result from the query.

So when you mention that both str and datetime can be used, should there be a check whether the type of the model field and matcher._value is same. If not, then convert it to the matcher._value type?

pubtools/pulplib/_impl/criteria.py
rajulkumar: Yes, that's how it will be post the above mentioned change.

pubtools/pulplib/_impl/fake/match.py
rajulkumar: Agree. Though this obj_value might need some conversion based on matcher._value type.

tests/client/test_search.py
rajulkumar: Agree. Should we consider int type fields for a similar conversion as well?(as a separate PR)

pubtools/pulplib/_impl/client/search.py
rohanpm: I am not sure that we actually need the str => datetime conversion.  Guess it would be a nice convenience, but you could consider first getting a patch without that and then letting it be added later.

If added, yes I think it'd work pretty similar to the way you just described.  Look up some converter function based on the model field's declared type, then try to apply it on the matcher._value.

tests/client/test_search.py
rohanpm: Hmm... I'm not really sure... I guess it's not as useful as the datetime conversion at least.

There's a reasonable argument for str => datetime conversion in that there is no supported datetime type across languages, and an ISO8601 string is really the de-facto standard for datetime interchange.  If a datetime is coming from a user, it probably was provided as an ISO8601 string in the first place.

For str => int, I guess it's not as useful, seems like it just enables the caller to be sloppy with their types, which could be a bad thing.

Needs more thought.  Either way, as you said let's not try to handle it in this PR.

pubtools/pulplib/_impl/client/search.py
```diff
@@ -79,12 +90,15 @@ def field_match(to_match):
         return {"$regex": to_match._pattern}
 
     if isinstance(to_match, EqMatcher):
-        return {"$eq": to_match._value}
+        return {"$eq": to_mongo_json(to_match._value)}
 
     if isinstance(to_match, InMatcher):
-        return {"$in": to_match._values}
+        return {"$in": to_mongo_json(to_match._values)}
```
rohanpm: I don't think this is correct, seems like it should rather be converting each value *in the list*.  For example, if you had a list containing:

```
['somestr', 123, datetime.datetime(...)]
```

Then I think you want the mongo conversion not to just return you the unmodified object but rather to return you:

```
['somestr', 123, {"$date": "..."}]
```

I suppose the correct place to handle this is to leave this code alone and do it inside of to_mongo_json, so I've made a suggestion there.

pubtools/pulplib/_impl/client/search.py
```diff
@@ -21,6 +23,15 @@ def all_subclasses(klass):
     return out
 
 
+def to_mongo_json(value):
+    # Return a value converted to the format expected for a mongo JSON
+    # expression. Only a handful of special types need explicit conversions.
+    if isinstance(value, datetime.datetime):
+        return {"$date": value.strftime("%Y-%m-%dT%H:%M:%SZ")}
+
```
rohanpm: Looks like this was oversimplified and should also recurse into lists and dicts, something like this:

```suggestion
    if isinstance(value, (list, tuple)):
        return [to_mongo_json(elem) for elem in value]
    if isinstance(value, dict):
        out = {}
        for (key, child_value) in value.items():
            out[key] = to_mongo_json(child_value)
        return out
```

So that e.g. testing if a value is $in one of a series of dates will work.

pubtools/pulplib/_impl/criteria.py
```diff
@@ -219,6 +219,28 @@ def in_(cls, values):
...
+
+        Arguments:
+            value (object)
+                An object to match against the field
+
+        Example:
+            .. code-block:: python
+
+            # would match where last_publish is before "2019-08-27T00:00:00Z"
+            # date comparison requries a dateime.datetime object
```
rohanpm:
```suggestion
            # date comparison requires a datetime.datetime object
```

pubtools/pulplib/_impl/criteria.py
```diff
@@ -219,6 +219,28 @@ def in_(cls, values):
...
+        value.
+
+        Arguments:
+            value (object)
+                An object to match against the field
+
+        Example:
+            .. code-block:: python
+
+            # would match where last_publish is before "2019-08-27T00:00:00Z"
```
rohanpm: If you build docs locally and review, you should see that this doesn't render correctly.
I think you need to indent everything you want to belong to the code-block.

pubtools/pulplib/_impl/fake/client.py
```diff
@@ -87,14 +87,34 @@ def search_repository(self, criteria=None):
...
+        criteria = criteria or Criteria.true()
+        distributors = []
+
+        filters_for_criteria(criteria, Distributor)
+
+        try:
+            for repo in self._repositories:
+                for distributor in repo.distributors:
+                    if match_object(criteria, distributor):
+                        distributors.append(distributor)
```
rohanpm: When using the real client, the distributor objects are always going to include a repo_id on the returned value.  But that's a new field, and any existing code using the fake client isn't going to be setting that, so the field could be missing on the returned values here.

Instead of returning the distributor directly, could we please return a copy of it with correct repo_id ensured?  e.g.

```suggestion
                        distributors.append(attr.evolve(distributor, repo_id=repo.id))
```

pubtools/pulplib/_impl/fake/match.py
```diff
@@ -124,6 +125,21 @@ def match_in(matcher, field, obj):
     return False
 
 
+@visit(LessThanMatcher)
+def match_less(matcher, field, obj):
+    # for datetime.dateime fields, obj_value is returned as pulp_value(iso format)
```
rohanpm:
```suggestion
    # for datetime.datetime fields, obj_value is returned as pulp_value(iso format)
```

But actually, I don't think it is right to have the datetime conversion embedded directly next to the less-than comparison here.  Is there a way to make it work for all of the comparisons like equals, in_ and so on?

pubtools/pulplib/_impl/model/distributor.py
```diff
@@ -26,6 +26,9 @@ class Distributor(PulpObject):
     of type `yum_distributor` may be used to create yum repositories.
     """
 
+    repo_id = pulp_attrib(type=str, default=None, pulp_field="repo_id")
```
rohanpm: With addition of this new attribute, I think a new validator in Repository class is appropriate to ensure the field is correct.

What I mean is, right now it's possible for bogus data to be created like this:

```
repo = Repository(id="some-id", distributors=(Distributor(id="dist-id", repo_id="oops-some-other-id"))
```

It would be best if such code raised an exception rather than allowing the invalid object to be created.  The validation would be something like: on each distributor, if repo_id is not None then it must match the ID of this repository.

Could you please look into that?

pubtools/pulplib/_impl/fake/match.py
rohanpm: I added https://github.com/release-engineering/pubtools-pulplib/pull/43/commits/bf154296612cc9809ad7452697eafc1df4004a50 onto your PR which I believe should fix this problem in a better way, could you please have a look?

rajulkumar: Thanks. This patch gets rid of that obj value vs matcher value inconsistencies.

pubtools/pulplib/_impl/model/distributor.py
rajulkumar: Sure. In the comments above, it's mentioned that distributor is always going to include repo_id. Should this exclude checking if repo_id is not None then compare. It should compare distributor.repo_id with repo.id in all cases.

rohanpm: I think the validation should be: for each Distributor on a Repository, if the distributor repo_id is not None, then it must be equal to the Repository's id.  So: yes, None should be tolerated.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -159,6 +159,20 @@ class Repository(PulpObject):
     _client = attr.ib(default=None, init=False, repr=False, cmp=False, hash=False)
     # hidden attribute for client attached to this object
 
+    @distributors.validator
+    def check_repo_id(self, _, value):
```
rohanpm: Should also be named with _ to make it private API, this method isn't meant to be invoked directly by users.

```suggestion
    def _check_repo_id(self, _, value):
```

rohanpm: I'd have committed this fix myself but it seems like I don't have permission (is the checkbox ticked to allow maintainers to edit this pull request?)

rohanpm: This is the last remaining issue I could see.

rajulkumar: Yes, it is checked.

PR #45, opened 2019-09-02
CI: add reverse-dependency testing for pubtools-pulp

rohanpm:
Don't allow incoming changes to pubtools-pulplib unless
pubtools-pulp tests also pass with the changes. This gives added
protection against regressions and backwards-incompatible changes.


rohanpm:
Results can be seen here: https://travis-ci.org/release-engineering/pubtools-pulplib/jobs/579598949

PR #48, opened 2019-09-03
Deprecate Page.as_iter in preference of __iter__

rohanpm:
Implementing `__iter__` allows to directly iterate over Page objects
rather than having to call as_iter.

The original idea with adding Page.as_iter rather than `__iter__` was
that using as_iter is blocking, and it would be better to be
explicit rather than implicit whenever operations might block, so
as to help keep a workflow entirely non-blocking.

However, now we have several pieces of code using this library and
the following pattern is seen commonly:

    repos = client.search_repository(...).result().as_iter()
    for repo in repos:
      ...

That's a bit too noisy, and this usage seems to be more common than
using search results in a non-blocking manner anyway (e.g. we usually
want to find & validate *all* repos before moving to the next step).

So, it seems like the idea of "explicit over implicit" here isn't
helping. Let's allow using the result as an iterable directly, so
above code can be written as slightly cleaner:

    repos = client.search_repository(...).result()
    for repo in repos:
      ...

### DIFF
--- a/pubtools/pulplib/_impl/page.py
+++ b/pubtools/pulplib/_impl/page.py
@@ -92,6 +92,7 @@ class Page(object):
             self.__dict__["_cancel_ref"] = weakref.ref(self, do_cancel)
 
     def as_iter(self):
+        # TODO: remove me. Originally deprecated 2019-09.
         warnings.warn(
             "as_iter is deprecated, use page as iterable instead", DeprecationWarning
         )
--- a/pubtools/pulplib/_impl/page.py
+++ b/pubtools/pulplib/_impl/page.py
@@ -1,5 +1,6 @@
 import logging
 import weakref
+import warnings
 
 from . import compat_attr as attr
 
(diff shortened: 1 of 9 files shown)


pubtools/pulplib/_impl/page.py
```diff
@@ -87,12 +92,13 @@ def do_cancel(*_):
             self.__dict__["_cancel_ref"] = weakref.ref(self, do_cancel)
 
     def as_iter(self):
-        """Returns an iterator which individually yields each object in this
-        page, and all subsequent pages.
```
midnightercz: Can we also add here some TODO, fox FIX or whatever keyword that could indicate that this should be removed at version x.y.z ?

rajulkumar: +1
Also, if planning to remove as_iter in later version, I would suggest to get that done sooner as this library has a relatively smaller usage across projects as of now. This will lead to minimal changes with backward incompatibility once as_iter is removed.

pubtools/pulplib/_impl/page.py
```diff
@@ -87,12 +92,13 @@ def do_cancel(*_):
...
+        warnings.warn(
```
rohanpm:
```suggestion
        # TODO: remove me. Originally deprecated 2019-09.
        warnings.warn(
```

pubtools/pulplib/_impl/page.py
rohanpm: I've added a TODO comment now.  I think removal of deprecated features should be time-based and not version-based (we discussed it on mailing list a while back, thread "Default deprecation period?"), so I record the approximate date at which it was deprecated rather than try to set now at which version it will be removed.

Issue #24, opened 2019-06-25
Improve docs for PulpObject and other attrs classes


rohanpm:
#### Summary

It would be nice if documentation of attributes on PulpObject subclasses, and other attrs-based classes, was improved with respect to the type and default value of attributes.

#### Issue Type
- Feature Request

#### Current behavior

If I look at generated docs e.g. at https://release-engineering.github.io/pubtools-pulplib/api/model.html, I see attribute docs like this:

> 
> is_sigstore = None
>  True if this is a sigstore repository, used for container image manifest signatures.
>
> signing_keys = None
>    A list of GPG signing key IDs used to sign content in this repository.
>

Where:

- the type of each attribute isn't clearly documented - it depends on the developer explicitly writing it in the doc string for the attribute (which can be easily forgotten)
- every single attribute has this trailing "= None", which is useless - it's displayed even for attributes which are mandatory to provide with a non-None value.

#### Desired behavior

Attribute docs don't have the useless "= None" suffix, and they display the attribute type in some standardized way.

For example:

> 
> is_sigstore: bool
>  True if this is a sigstore repository, used for container image manifest signatures.
>
> signing_keys: List[str]
>    GPG signing key IDs used to sign content in this repository.
>

#### Use case

Allows developers to more clearly understand expected type of all attributes, and cuts down on useless noise in the docs.

PR #49, opened 2019-09-04
Refactor models to satisfy desired invariants

rohanpm:
Our model classes had various bugs and weaknesses.  This pull request
fixes them and adds tests to ensure that the same mistakes can't happen
on new classes.

The original motivation here was to fix up inconsistent usage of tuples vs lists
for immutability, but testing found several additional bugs to be fixed, most
notably that the recently added maintenance classes had fields using an
entirely different type than what was declared in the model.

The changes made here include:

- added tests which verify that *all* public attrs-using classes meet certain invariants,
  such as "actual types match declared types" and "all fields are immutable"
- added a FrozenList immutable list type and used that for all model fields which were
  formerly using tuples or plain lists
- fix incorrect implementations of Maintenance classes which used wrong types
- make Page immutable too (seems like an oversight that it wasn't already)
- bump version to 2.0.0 because these changes are not strictly backwards-compatible
  (although I doubt that any current users are affected).

More information on each of these changes can be found in the commit messages.

### DIFF
--- a/pubtools/pulplib/_impl/model/frozenlist.py
+++ b/pubtools/pulplib/_impl/model/frozenlist.py
@@ -1,0 +1,35 @@
+class FrozenList(list):
+    # An immutable list subclass, intended for use on model fields.
+
+    # Set of overridden methods is taken from collections.abc.MutableSequence.
+
+    def __raise_immutable(self):
+        raise NotImplementedError("cannot modify immutable list")
+
+    def __setitem__(self, *_args, **_kwargs):
+        self.__raise_immutable()
+
+    def __delitem__(self, *_args, **_kwargs):
+        self.__raise_immutable()
+
+    def __iadd__(self, *_args, **_kwargs):
+        self.__raise_immutable()
+
+    def insert(self, *_args, **_kwargs):
+        self.__raise_immutable()
+
+    def append(self, *_args, **_kwargs):
+        self.__raise_immutable()
+
+    def extend(self, *_args, **_kwargs):
+        self.__raise_immutable()
+
+    def pop(self, *_args, **_kwargs):
+        self.__raise_immutable()
+
+    def remove(self, *_args, **_kwargs):
+        self.__raise_immutable()
+
+    # FrozenList is hashable if everything within it is hashable
+    def __hash__(self):
+        return hash(tuple(self))
--- a/pubtools/pulplib/_impl/model/repository/base.py
+++ b/pubtools/pulplib/_impl/model/repository/base.py
@@ -6,6 +6,7 @@ from more_executors.futures import f_map
 from ..common import PulpObject, DetachedException
 from ..attr import pulp_attrib
 from ..distributor import Distributor
+from ..frozenlist import FrozenList
 from ...schema import load_schema
 from ... import compat_attr as attr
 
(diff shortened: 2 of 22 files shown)


pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -89,14 +91,15 @@ class Repository(PulpObject):
     """
 
     distributors = pulp_attrib(
-        default=attr.Factory(tuple),
-        type=tuple,
+        default=attr.Factory(FrozenList),
+        type=list,
```
midnightercz: Why not "FrozenList"?
In docs there's:

> Please note that attrs doesn’t do anything with this metadata by itself. You can use it as part of your own code or for static type checking.

having list here makes it confusing

pubtools/pulplib/_impl/model/frozenlist.py
```diff
@@ -0,0 +1,35 @@
+class FrozenList(list):
```
midnightercz: What the benefit of having custom class instead of tuple?

JayZ12138: I just checked the doc of atrs, for 'default' value for an attr.ib, it says:
`If the value is an instance of Factory, its callable will be used to construct a new value (useful for mutable data types like lists or dicts).`

So, if it's a Frozen list, which inherits from list, then a new FronzenList object will be constructed, but if it's a tuple, it won't.

rohanpm: I wrote explanation for that in https://github.com/release-engineering/pubtools-pulplib/pull/49/commits/75790f140edc0f4fe1a330bcf407cee8fcce0e37 commit message.  In summary: the previous solution *was* to use tuple but it has usability issues because developers choose lists as the default container and tuples aren't as compatible with them as it may seem at first glance.  For instance `(1, 2, 3) == [1, 2, 3]` isn't true.

That means, if it continued using tuples, developers have to get in the habit of remembering where they need to write a tuple and where they need to write a list, or writing tuples everywhere (which is a tough ask), or running into weird bugs.

pubtools/pulplib/_impl/model/repository/base.py
rohanpm: The attrs docs aren't quite accurate, or at least, aren't telling the whole story.  These types do actually make it into the API as they are added as Python3 type annotations, where tools like pylint, mypy and sphinx look at them.  I expect them to end up in our docs eventually, though it's not quite working right now (https://github.com/release-engineering/pubtools-pulplib/issues/24).

For example, before this change the declared type of 'distributors' argument shows up in API as:

```
>>> inspect.signature(Repository.__init__).parameters['distributors'].annotation
<class 'tuple'>
```

After this change, it will be 'list'.  I do believe list is more appropriate than FrozenList here because:

- setting `type=FrozenList` is exposing that type as part of our public API, but I don't want to commit to returning instances of exactly that class
- given that it's used as type annotation on the constructor, the appropriate usage seems more like 'declare this as the type you *accept*' and not 'declare this as the type you *return*/*store*'.  So, type=list is fine as the constructor does indeed accept *any* type of list and doesn't have to be passed exactly a FrozenList.

PR #54, opened 2019-09-11
User should be able to search distributors by relative_url

JayZ12138:
Add a new attribute, relative_url, to Distributor, which makes it
possible to search distributors by relative_url.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- n/a
+### Added
+- Added a new attribute a`relative_url` to `Distributor`, so users can search distributors
+  by relative_url
 
 ## [2.1.0] - 2019-09-10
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- n/a
+### Added
+- Added a new attribute a`relative_url` to `Distributor`, so users can search distributors
```
rohanpm:
```suggestion
- Added a new attribute `relative_url` to `Distributor`, so users can search distributors
```

tests/client/test_client.py
```diff
@@ -68,18 +69,71 @@ def test_can_search_distributor(client, requests_mocker):
...
+            },
+            {
+                "id": "cdn_distributor",
+                "distributor_type_id": "rpm_rsync_distributor",
+                "config": {"relative_url": "relative/path"},
+            },
+        ],
+    )
+
+    crit = Criteria.with_field("relative_url", Matcher.regex("relative/path"))
```
rohanpm: This test would have been a little stronger if the test data and matcher were set up so this actually had to be a regex comparison.  Right now the test would pass if the regex was not evaluated as a regex but just an exact string comparison, since exactly the same values were chosen for the pattern and for the data being tested.

PR #66, opened 2019-09-30
minor bug fix and rewrite set-maintenance example script

JayZ12138:
Now the set-maintenance script example takes the --repo-url-regex
option and set maintenance mode on based on the relative_url, instead
of id when this option is used.

Also, fix a minor bug get_maintenance map_404_to_non callback, which
the response could be None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- Fixed minor bug in map_404_to_none, where the response could have response equals to None
-- rewrite set-maintenace example, make it more realistic
+- Fixed certain exceptions from requests library not being propagated properly while
+  getting maintenance report.
 
 ## [2.3.0] - 2019-09-25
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- n/a
+### Fixed
+- Fixed minor bug in map_404_to_none, where the response could have response equals to None
```
rohanpm: I don't think this kind of changelog entry makes sense, because there is no map_404_to_none in the API.  Can you please describe this in terms that an end-user of the library could understand?  How did this bug affect any code using the library?

CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- rewrite set-maintenace example, make it more realistic
```
rohanpm: I don't think example changes need to be mentioned in changelog at all, because they're not included in releases.  (Examples don't go into the source tarball uploaded to pypi.)

CHANGELOG.md
rohanpm: And since https://github.com/release-engineering/pubtools-pulplib/pull/51 was fixing the same bug in a different piece of code, I guess the commit message / changelog entry there includes some hints.

PR #67, opened 2019-10-01
Fix handling of errors with response=None

rohanpm:
Our retry policy has custom handling for 404 responses.
The code for that had wrongly assumed that, whenever an exception
had a response attribute, that attribute would be something
other than None.

In practice, the response is None for certain exceptions generated
by the requests library where a failure occurred before a response
was available (e.g. SSL handshake errors, I/O errors). Make sure
we tolerate this, so those exceptions propagate correctly rather
than being transformed to an internal error in the retry policy.


rohanpm:
Mistaken repush of something already merged.

PR #68, opened 2019-10-01
Ensure Matcher.regex raises when passed non-string

rohanpm:
Matcher.regex is documented as accepting a string, but previously
if you passed an already-compiled regex, that would also be
accepted by the validator. This happened because re.compile simply
returns already-compiled regexes without an error. It would
ultimately lead to an error when executing a search, because a
compiled regex is not serializable to JSON.

Fix it to more strictly typecheck the inputs when creating the
matcher, so we fail early and with a meaningful message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- n/a
+### Changed
+- Patterns to `Matcher.regex` are now more strictly typechecked when the matcher is created.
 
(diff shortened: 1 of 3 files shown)


rohanpm:
CI is failing as the reverse-dependency test is showing that pubtools-pulp tests would fail with this change.  That's expected until https://github.com/release-engineering/pubtools-pulp/pull/15 is submitted.  The issue is that pubtools-pulp currently relies on this missing validation for its own autotests.

So, this is fine for review now, but https://github.com/release-engineering/pubtools-pulp/pull/15 must be submitted and then this should be retested before merging.

Issue #64, opened 2019-09-30
Use *Unit classes as templates in search criteria


rbikar:
#### Summary

To avoid using raw dicts searching in Pulp we could use *Unit classes as templates in search criteria.
So we can introduce a class method Criteria.for_units() that will accept list of units for which   criteria should be created, unset fields in unit instance will be interpreted as "match any value for this field"
E.g. we can do this for copying or removing or searching units:
```
unit = RpmUnit(name='bash')
crit = Criteria.for_units([unit])
# This kind of criteria can be used in
# various APIs like:
# copy all bash from repo1 to repo2
repo1.copy_content(criteria=crit, to_repo=repo2)
# remove all bash RPMs
repo1.remove_content(criteria=crit)
# search all bash RPMs
repo1.search_content(criteria=crit)
```
This should copy all RPMs named "bash" from one repo to another, or it should search in similar manner.

#### Issue Type
- Feature Request

#### Current behavior

Currently creation Criteria instance for units is no much convenient.

#### Desired behavior

- Lib supports creation of criteria with *Unit classes used as templates (Criteria.for_units([unit1, unit2]))
- Unset filed in unit are interpreted as match any value for this field

PR #71, opened 2019-11-11
Introduce search_content method for repositories [DELIVERY-7839]

nathanegillett:
This commit adds the search_content method for the Repository class
which searches for and retrieves content units from Pulp repositories.

### DIFF
--- a/examples/search-repo-content
+++ b/examples/search-repo-content
@@ -46,6 +46,17 @@ def main():
     client = make_client(p)
     units = client.search_repository_content(p.repo_id, type_ids=p.content_type)
 
+    """
+    Alternatively, one could use the following:
+
+    .. code-block:: python
+        repository = client.get_repository(p.repo_id)
+        units = repository.search_content(type_id=p.content_type)
+
+    This is functionally equivalent to client.search_repository_content
+    but is useful if further operations on the repository are required.
+    """
+
     if units:
         log.info("Found %s %s units: \n\t%s" % (
             len(units), p.content_type,
--- a/examples/search-repo-content
+++ b/examples/search-repo-content
@@ -3,9 +3,9 @@ import os
 import logging
 from argparse import ArgumentParser
 
-from pubtools.pulplib import Client, Criteria
+from pubtools.pulplib import Client
 
-log = logging.getLogger("repository-content-search")
+log = logging.getLogger("search-repo-content")
 
 
 def make_client(args):
@@ -26,7 +26,9 @@ def main():
     log.setLevel(logging.INFO)
     logging.basicConfig(format="%(message)s", level=logging.INFO)
 
-    parser = ArgumentParser(description="Retrieve unit objects of the specified type from a repository")
+    parser = ArgumentParser(
+        description="Retrieve unit objects of the given type from the given repository"
+    )
     parser.add_argument("--url", help="Pulp server URL", required=True)
     parser.add_argument("--username", help="Pulp username")
     parser.add_argument("--password", help="Pulp password (or set PULP_PASSWORD in env)")
@@ -42,8 +44,16 @@ def main():
         log.setLevel(logging.DEBUG)
 
     client = make_client(p)
-    criteria = Criteria.with_field("type_ids", [p.content_type])
-    units = client.search_repository_content(p.repo_id, criteria)
+    units = client.search_repository_content(p.repo_id, type_ids=p.content_type)
+
+    if units:
+        log.info("Found %s %s units: \n\t%s" % (
+            len(units), p.content_type,
+            "\n\t".join(str(unit) for unit in units)
+        ))
+    else:
+        log.info("No %s units found." % p.content_type)
+
     return units
 
 
--- a/examples/search-repo-content
+++ b/examples/search-repo-content
@@ -44,18 +44,8 @@ def main():
         log.setLevel(logging.DEBUG)
 
     client = make_client(p)
-    units = client.search_repository_content(p.repo_id, type_ids=p.content_type)
-
-    """
-    Alternatively, one could use the following:
-
-    .. code-block:: python
-        repository = client.get_repository(p.repo_id)
-        units = repository.search_content(type_id=p.content_type)
-
-    This is functionally equivalent to client.search_repository_content
-    but is useful if further operations on the repository are required.
-    """
+    repository = client.get_repository(p.repo_id)
+    units = repository.search_content(type_id=p.content_type)
 
     if units:
         log.info("Found %s %s units: \n\t%s" % (
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -208,9 +208,9 @@ class Client(object):
 
         Args:
             repository_id (str)
-                The ID of the repository in which to search
+                The ID of the repository to search.
             type_ids (str, list, tuple)
-                A list of content types to search
+                A list of content types to search.
             criteria (:class:`~pubtools.pulplib.Criteria`)
                 A criteria object used for this search.
 
@@ -225,13 +225,13 @@ class Client(object):
 
         .. versionadded:: 2.4.0
         """
-        search_f = self._search(
-            Unit,
-            "repositories/%s" % repository_id,
-            criteria=criteria,
-            type_ids=validate_type_ids(type_ids),
+        resource_type = "repositories/%s" % repository_id
+        search_options = {"type_ids": validate_type_ids(type_ids)}
+        return list(
+            self._search(
+                Unit, resource_type, criteria=criteria, search_options=search_options
+            )
         )
-        return [unit for unit in search_f]
 
     def search_distributor(self, criteria=None):
         """Search the distributors matching the given criteria.
@@ -252,14 +252,8 @@ class Client(object):
         """
         return self._search(Distributor, "distributors", criteria=criteria)
 
-    def _search(
-        self,
-        return_type,
-        resource_type,
-        criteria=None,
-        search_options=None,
-        type_ids=None,
-    ):
+    def _search(self, return_type, resource_type, criteria=None, search_options=None):
+        type_ids = search_options.pop("type_ids", None) if search_options else None
         pulp_crit = {
             "skip": 0,
             "limit": self._PAGE_SIZE,
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -231,7 +231,7 @@ class Client(object):
             criteria=criteria,
             type_ids=validate_type_ids(type_ids),
         )
-        return [unit for unit in search_f.result().as_iter()]
+        return [unit for unit in search_f]
 
     def search_distributor(self, criteria=None):
         """Search the distributors matching the given criteria.
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -203,11 +203,11 @@ class Client(object):
             Repository, "repositories", criteria=criteria, search_options=search_options
         )
 
-    def search_repository_content(self, repository_id, type_ids, criteria=None):
+    def search_repository_content(self, repo_id, type_ids, criteria=None):
         """Search the given repository for content matching the given criteria.
 
         Args:
-            repository_id (str)
+            repo_id (str)
                 The ID of the repository to search.
             type_ids (str, list, tuple)
                 A list of content types to search.
@@ -225,7 +225,7 @@ class Client(object):
 
         .. versionadded:: 2.4.0
         """
-        resource_type = "repositories/%s" % repository_id
+        resource_type = "repositories/%s" % repo_id
         search_options = {"type_ids": validate_type_ids(type_ids)}
         return list(
             self._search(
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -13,7 +13,7 @@ from more_executors.futures import f_map, f_flat_map, f_return, f_proxy
 
 from ..page import Page
 from ..criteria import Criteria
-from ..model import Repository, MaintenanceReport, Distributor
+from ..model import Repository, MaintenanceReport, Distributor, Unit
 from .search import filters_for_criteria
 from .errors import PulpException
 from .poller import TaskPoller
@@ -231,13 +231,18 @@ class Client(object):
         search = {"criteria": pulp_crit}
         search.update(search_options or {})
 
-        response_f = self._do_search(resource_type, search)
+        url = os.path.join(self._url, "pulp/api/v2/%s/search/" % resource_type)
+
+        if return_type is Unit and "repositories/" in resource_type:
+            url = os.path.join(url, "units/")
+
+        response_f = self._do_search(url, search)
 
         # When this request is resolved, we'll have the first page of data.
         # We'll need to convert that into a page and also keep going with
         # the search if there's more to be done.
         return f_proxy(
-            f_map(response_f, lambda data: self._handle_page(return_type, search, data))
+            f_map(response_f, lambda data: self._handle_page(url, return_type, search, data))
         )
 
     def get_maintenance_report(self):
@@ -417,12 +422,14 @@ class Client(object):
             pass
         return taskdata
 
-    def _handle_page(self, object_class, search, raw_data):
+    def _handle_page(self, url, object_class, search, raw_data):
         LOG.debug("Got pulp response for %s, %s elems", search, len(raw_data))
 
         page_data = [object_class.from_data(elem) for elem in raw_data]
         for obj in page_data:
-            obj._set_client(self)
+            # Set client is only applicable for repository and distributor objects
+            if hasattr(obj, "_set_client"):
+                obj._set_client(self)
 
         # Do we need a next page?
         next_page = None
@@ -433,11 +440,11 @@ class Client(object):
             search = search.copy()
             search["criteria"] = search["criteria"].copy()
             search["criteria"]["skip"] = search["criteria"]["skip"] + limit
-            response_f = self._do_search("repositories", search)
+            response_f = self._do_search(url, search)
             next_page = f_proxy(
                 f_map(
                     response_f,
-                    lambda data: self._handle_page(object_class, search, data),
+                    lambda data: self._handle_page(url, object_class, search, data),
                 )
             )
 
@@ -458,9 +465,7 @@ class Client(object):
     def _do_request(self, **kwargs):
         return self._session.request(**kwargs)
 
-    def _do_search(self, resource_type, search):
-        url = os.path.join(self._url, "pulp/api/v2/{0}/search/".format(resource_type))
-
+    def _do_search(self, url, search):
         LOG.debug("Submitting %s search: %s", url, search)
         return self._request_executor.submit(
             self._do_request, method="POST", url=url, json=search
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -242,7 +242,10 @@ class Client(object):
         # We'll need to convert that into a page and also keep going with
         # the search if there's more to be done.
         return f_proxy(
-            f_map(response_f, lambda data: self._handle_page(url, return_type, search, data))
+            f_map(
+                response_f,
+                lambda data: self._handle_page(url, return_type, search, data),
+            )
         )
 
     def get_maintenance_report(self):
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -12,7 +12,7 @@ from more_executors import Executors
 from more_executors.futures import f_map, f_flat_map, f_return, f_proxy
 
 from ..page import Page
-from ..criteria import Criteria, AndCriteria, FieldMatchCriteria
+from ..criteria import Criteria, AndCriteria, OrCriteria, FieldMatchCriteria
 from ..model import Repository, MaintenanceReport, Distributor, Unit
 from ..model.common import InvalidContentTypeException
 from ..model.unit import SUPPORTED_UNIT_TYPES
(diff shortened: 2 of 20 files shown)


nathanegillett:
Not sure why it's complaining about coverage. The tests included do cover these lines.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -188,6 +190,57 @@ def distributor(self, distributor_id):
         """
         return self._distributors_by_id.get(distributor_id)
 
+    def search_content(self, name=None, arch=None, filename=None, stream=None):
```
rohanpm: It shouldn't have the unit-specific fields like name, arch, filename, stream directly built-in to the method signature like this.  It won't scale to support all unit types and it's entirely inconsistent with how the existing search methods work (like searching for repositories).  It should take the criteria object directly, rather than only allowing the method to be used for some specific hardcoded cases.

type_ids also shouldn't be hardcoded in the method body.  This should work with all the unit types supported by the library, so it should work for ISO units too.  Plus, there's a performance penalty which may be significant when you search for type_ids you don't need.  As I understand it, each type_id results in an additional mongo search (if the repo contains any units of that type).  I think the caller needs to specify which types of unit they're interested in.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -188,6 +190,57 @@ def distributor(self, distributor_id):
...
+                Name of the desired unit.
+            arch (str)
+                Architecture of the desired unit.
+            filename (str)
+                File name of the desired unit.
+            stream (str)
+                Stream of the desired modulemd or modulemd_defaults unit.
+
+        Returns:
+            Future[list[:class:`~pubtools.pulplib.Task`]]
```
rohanpm: This doesn't seem to be accurate.

pubtools/pulplib/_impl/schema/unit.yaml
```diff
@@ -95,7 +95,9 @@ definitions:
     properties:
       # Type of the unit
       _content_type_id:
-        const: modulemd
```
rohanpm: This doesn't seem to be right at all - you're letting modulemd and modulemd_defaults share the same schema here, but they have a different set of expected fields.

For instance, modulemd_defaults doesn't have a "context" field, yet it's listed here as a required field.  I'd expect this not to work.  If I'm right then it means some autotests are also missing.

tests/repository/test_search_content.py
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    "_content_type_id": "modulemd",
+                    "name": "m1",
+                    "stream": "s1",
+                    "version": 1234,
+                    "context": "a1b2c3",
+                    "arch": "s390x",
+                },
+                {
+                    "_content_type_id": "modulemd_defaults",
```
rohanpm: This isn't what a real modulemd_defaults unit looks like.  If you put in proper test data here, you should see a crash in schema checking (see earlier comment about wrong schema).

nathanegillett:
> I wonder if we should merge this and #64 together.
I think I'd rather see #64 replace the criteria implemented here with unit models rather than include all the work necessary to make that work in this issue.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -188,6 +190,57 @@ def distributor(self, distributor_id):
...
+
+        .. versionadded:: 2.4.0
+        """
+        if not self._client:
+            raise DetachedException()
+
+        criteria = (
+            Criteria.with_field(
+                "type_ids",
+                Matcher.in_(["rpm", "srpm", "modulemd", "modulemd_defaults"]),
```
midnightercz: I think what you should do here is something like this:
In pubtools/pulplib/_impl/model/unit/__init__.py  add something like SUPPORTED_UNITS = (FileUnit, RpmUnit,....) and also make this constant public.
and here you sould basically accept type_ids in search and check if type_ids are subset of SUPPORTED_UNITS.content_type_id - or accept list of Unit* classes directly as type_ids attribute.
More dynamic way would be to check for content of the module and grep for all classes with base "Unit" and then accepts type_ids based on that.

pubtools/pulplib/_impl/schema/unit.yaml
nathanegillett: Yeah, I should've checked this instead of assuming they had the same schema. Fixed in next commit.

pubtools/pulplib/_impl/model/repository/base.py
nathanegillett: Good ideas for validating the content type id. I'll be sure it's included.

nathanegillett: "if type_id not in [cls.__name__ for cls in Unit.__subclasses__()]" should do it.

nathanegillett:
> I think you're missing support on the fake client - if you have a Repository obtained from a fake client now and you call search_content, what happens? That should work too & would need a test.

search_content belongs to Repository, though, and uses client._search for which there's no fake method.

nathanegillett:
> I think you're missing support on the fake client - if you have a Repository obtained from a fake client now and you call search_content, what happens? That should work too & would need a test.

The repository object is still real, right? And, though it makes sense that the real client isn't being called, at least search_content should be covered as it belongs to Repository.
I'm not understanding how it isn't being reported as covered or how client._search was previously covered if only the fake client is used in tests.

rohanpm:
> > I think you're missing support on the fake client - if you have a Repository obtained from a fake client now and you call search_content, what happens? That should work too & would need a test.
> 
> The repository object is still real, right? And, though it makes sense that the real client isn't being called, at least search_content should be covered as it belongs to Repository.

The point is that if you use a fake client to get a Repository object, then call search_content, it must behave in a way which makes sense, i.e. it returns a consistent view of whatever data is stored on the fake client.  Does it behave that way now?  I'd expect it's necessary to write new code to achieve that, and then that new code will need to be tested.

> I'm not understanding how it isn't being reported as covered or how client._search was previously covered if only the fake client is used in tests.

It's not the case that only the fake client is used in tests.

This library provides two implementations of a Pulp client for use by others, one designed to issue real HTTP requests to a Pulp server and the other designed to return data from local memory.  Both implementations should be tested by this library's test suite, so there are some tests here using one and some tests using the other.

> Also, apparently all the Unit subclasses were covered but I see no tests for them. So, I'm not sure how to get ModulemdDefaultUnit covered either.

Autotest tests/model/test_model_invariants.py automatically finds and performs some basic tests on every model class.

On top of that, at least one test where a search returns a ModulemdDefaultUnit would be nice, which I think you already have added.  I don't think there's much need for adding more tests when introducing new unit types.

nathanegillett:
> It's not the case that only the fake client is used in tests.
> [...]
> Autotest tests/model/test_model_invariants.py automatically finds and performs some basic tests on every model class.

The main problem that I'm having is that coveralls is claiming that nothing I've implemented is covered, and that just doesn't seem possible based on the tests I've included and what you've said here.

pubtools/pulplib/_impl/schema/unit.yaml
```diff
@@ -117,6 +117,33 @@ definitions:
...
+        type: string
+
+      # Repository associated with the unit
+      repo_id:
+        type: string
+
+      # Stream of the unit
+      stream:
+        type: string
+
```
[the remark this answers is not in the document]
nathanegillett: Will do.

pubtools/pulplib/_impl/model/unit/modulemd_defaults.py
```diff
@@ -0,0 +1,32 @@
...
+from ..attr import pulp_attrib
+from ... import compat_attr as attr
+
+
+@unit_type("modulemd")
+@attr.s(kw_only=True, frozen=True)
+class ModulemdDefaultsUnit(Unit):
+    """A :class:`~pubtools.pulplib.Unit` representing a modulemd_defaults document.
+
+    .. versionadded:: 1.5.0
```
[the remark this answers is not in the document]
nathanegillett: Fixed

pubtools/pulplib/_impl/model/unit/modulemd_defaults.py
```diff
@@ -0,0 +1,32 @@
+from .base import Unit, unit_type
+
+from ..attr import pulp_attrib
+from ... import compat_attr as attr
+
+
+@unit_type("modulemd")
```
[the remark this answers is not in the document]
nathanegillett: Fixed

pubtools/pulplib/_impl/model/unit/modulemd_defaults.py
```diff
@@ -0,0 +1,32 @@
...
+class ModulemdDefaultsUnit(Unit):
+    """A :class:`~pubtools.pulplib.Unit` representing a modulemd_defaults document.
+
+    .. versionadded:: 1.5.0
+    """
+
+    name = pulp_attrib(type=str, pulp_field="name")
+    """The name of this modulemd defaults unit.
+
+    Example: the name of ant:rhel-8-for-aarch64-appstream-htb-rpms is "ant".
```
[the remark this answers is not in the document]
nathanegillett: Fixed

tests/repository/test_search_content.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class TestSearchContent(object):
+    def __init__(self, requests_mocker, client):
+        self.repo = Repository(id="some-repo")
+        self.repo.__dict__["_client"] = client
+
+        requests_mocker.post(
+            "https://pulp.example.com/pulp/api/v2/repositories/some-repo/search/units/",
+            json=[
```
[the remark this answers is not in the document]
nathanegillett: Thanks for pointing this out

pubtools/pulplib/_impl/model/unit/modulemd_defaults.py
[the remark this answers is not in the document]
nathanegillett: Fixed

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -188,6 +195,42 @@ def distributor(self, distributor_id):
         """
         return self._distributors_by_id.get(distributor_id)
 
+    def search_content(self, type_id, criteria=None):
```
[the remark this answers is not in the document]
nathanegillett: Done

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -188,6 +195,42 @@ def distributor(self, distributor_id):
...
+
+        .. versionadded:: 2.4.0
+        """
+        if not self._client:
+            raise DetachedException()
+
+        if type_id not in [cls.__name__ for cls in Unit.__subclasses__()]:
+            raise InvalidContentTypeException()
+
+        type_crit = Criteria.with_field("_content_type_ids", Matcher.in_([type_id]))
```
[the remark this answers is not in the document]
nathanegillett: Fixed

rohanpm:
> 
> The main problem that I'm having is that coveralls is claiming that nothing I've implemented is covered, and that just doesn't seem possible based on the tests I've included and what you've said here.

I can reproduce the coveralls results locally.

- modulemd_defaults.py isn't tested because it's never imported.  It also means that it's not covered  by the test_model_invariants.py test I mentioned - a class has to at least be present in the library's public API for it to be found & tested.  It also won't appear in the documentation.  You can handle it similarly to ModulemdUnit to fix these problems.
- Although you do have a test which will load data of type modulemd_defaults, you never verify that the resulting object is an instance of ModulemdDefaultsUnit, so I think the library is falling into the default case it takes for all units of unknown type, which is to return an instance of the Unit base class.  If you add at least one test which can only pass if the returned unit is genuinely a ModulemdDefaultsUnit, it would protect against this bug.
- The test class you added, TestSearchContent, isn't being run at all.  I can see this from the output of py.test (need -v option to list each test function run).  If you want to make a test *class*, you apparently need to subclass unittest.TestCase, which hasn't been done here.  But I would rather recommend following the same style as existing tests in this repo, using bare functions named test_* and not unnecessarily wrapped into a class.

rohanpm:
> I can reproduce the coveralls results locally.
> (... 3 points)

Points 2 and 3 are probably conflicting with each other a bit, as I was reviewing the TestSearchContent body with the understanding that it would be executed, which later turned out to be false.  If you fix point 3 first, it might fix or partially fix point 2.

nathanegillett:
> [...] coveralls results [...]

Yeah, I figured out a lot of this myself last night and think I've solved most of the points you mention.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -188,6 +195,42 @@ def distributor(self, distributor_id):
...
+    def search_content(self, type_id, criteria=None):
+        """Search this repository for content matching the given criteria.
+
+        Args:
+            type_id (str)
+                Type of the desired unit.
+            criteria (:class:`~pubtools.pulplib.Criteria`)
+
+        Returns:
+            Future[list[:class:`~pubtools.pulplib.Unit`]]
```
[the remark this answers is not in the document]
nathanegillett: This actually returned subclasses of Unit, depending on the search issued.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -188,6 +195,40 @@ def distributor(self, distributor_id):
...
+        return self._search_content(Criteria.with_field("type_ids", ["modulemd"]))
+
+    @property
+    def modulemd_defaults_content(self):
+        """A list of modulemd_defaults units stored in this repository"""
+        return self._search_content(
+            Criteria.with_field("type_ids", ["modulemd_defaults"])
+        )
+
+    def _search_content(self, criteria=None):
```
nathanegillett: This shouldn't be private. Will fix in next commit.

nathanegillett:
Is anyone (@release-engineering/content-delivery) able to see why criteria are now broken?

nathanegillett:
> Is anyone (@release-engineering/content-delivery) able to see why criteria are now broken?

I'll add that this is true even against real pulp servers via examples/repository-content-search. The return is every unit in the given repo regardless of the given content_type.

tests/fake/test_fake_search.py
```diff
@@ -206,8 +213,12 @@ def test_search_bad_criteria():
     with pytest.raises(Exception) as dist_exc:
         client.search_distributor("invalid criteria")
 
+    with pytest.raises(Exception) as repo_cont_exc:
+        client.search_repository_content("some-repo", "rpm", "criteria?")
+
     assert "Not a criteria" in str(repo_exc.value)
     assert "Not a criteria" in str(dist_exc.value)
+    assert "Not a criteria" in str(repo_cont_exc.value)
```
nathanegillett: ...while this is true?

pubtools/pulplib/_impl/fake/client.py
```diff
@@ -90,6 +91,24 @@ def search_repository(self, criteria=None):
...
+        filters_for_criteria(criteria, Unit)
+
+        try:
+            for unit in self._repo_units[repository_id]:
+                if unit.content_type_id in validate_type_ids(type_ids) and match_object(
+                    criteria, unit
+                ):
+                    units.append(attr.evolve(unit))
+        except Exception as ex:  # pylint: disable=broad-except
+            return f_return_error(ex)
```
nathanegillett: How is this not covered...

nathanegillett: How is this not covered...

tests/fake/test_fake_search.py
nathanegillett: ...while this is true?

nathanegillett:
Finally.
Please review, @release-engineering/content-delivery.

pubtools/pulplib/_impl/client/search.py
```diff
@@ -113,3 +117,34 @@ def field_match(to_match):
...
+
+
+def validate_type_ids(type_ids):
+    valid_type_ids = []
+    invalid_type_ids = []
+
+    if isinstance(type_ids, str):
+        type_ids = [type_ids]
+
+    if not isinstance(type_ids, (list, tuple)):
```
midnightercz: minor, but I think you basically what to know if hasattr(type_ids, "\_\_iter\_\_")

pubtools/pulplib/_impl/client/search.py
```diff
@@ -113,3 +117,34 @@ def field_match(to_match):
         return {"$lt": to_mongo_json(to_match._value)}
 
     raise TypeError("Not a matcher: %s" % repr(to_match))
+
+
+def validate_type_ids(type_ids):
```
midnightercz: maybe util.py would be better module. I think this could be reused in different parts of the code

pubtools/pulplib/_impl/client/search.py
nathanegillett: Not sure that we want or need to open this up to other iterators, though.

pubtools/pulplib/_impl/client/search.py
nathanegillett: This started out a "filter", so it seemed to fit, but became what it is now, which, I agree, would probably be more at home in utils.py.

nathanegillett: Moving this throws a lot else out of whack. I'll keep trying but I may end up leaving it in search.py.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -203,6 +203,36 @@ def search_repository(self, criteria=None):
             Repository, "repositories", criteria=criteria, search_options=search_options
         )
 
+    def search_repository_content(self, repository_id, type_ids, criteria=None):
+        """Search the given repository for content matching the given criteria.
+
+        Args:
+            repository_id (str)
+                The ID of the repository in which to search
```
rajulkumar: minor: may be something like "repository ID to search"

nathanegillett: Well, that would make sense but we're searching for content **within** the repo with the given ID.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -203,6 +203,36 @@ def search_repository(self, criteria=None):
...
+
+        .. versionadded:: 2.4.0
+        """
+        search_f = self._search(
+            Unit,
+            "repositories/%s" % repository_id,
+            criteria=criteria,
+            type_ids=validate_type_ids(type_ids),
+        )
+        return [unit for unit in search_f.result().as_iter()]
```
rajulkumar: .result() and .as_iter() are not required as these are implicit in f_proxy and Page classes. So this can be [unit for unit in search_f]

nathanegillett: Oh, cool! Thanks!

pubtools/pulplib/_impl/client/client.py
rajulkumar: either is fine as long as it's understood. just thought too many linkers in the statement.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -203,6 +203,36 @@ def search_repository(self, criteria=None):
             Repository, "repositories", criteria=criteria, search_options=search_options
         )
 
+    def search_repository_content(self, repository_id, type_ids, criteria=None):
```
JayZ12138: so, you mainly exposed two public APIs, one is Client.search_repository_content and another is Repository.search_content and they basically did the same thing, am I right?
Do you think it's kinda redundant? Like publish, we only exposed it in Repository, not Client, since this operation is about repository directly.

nathanegillett: Right, you can either use Client.search_repository_content() to look up a repo's units or you can call search_content() from a repository object. I don't see the harm in leaving Client's method public -- it may help in some, probably infrequent, use cases. We can hide it, though, if there's reason to.

pubtools/pulplib/_impl/model/unit/modulemd_defaults.py
```diff
@@ -0,0 +1,28 @@
...
+    .. versionadded:: 2.4.0
+    """
+
+    name = pulp_attrib(type=str, pulp_field="name")
+    """The name of this modulemd defaults unit"""
+
+    stream = pulp_attrib(type=str, pulp_field="stream")
+    """The stream of this modulemd defaults unit"""
+
+    repository_id = pulp_attrib(type=str, pulp_field="repo_id")
```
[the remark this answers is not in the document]
nathanegillett: I think it's the opposite, actually -- this library seems to be pushing "repository_id", whereas I've only seen "repo_id" when dealing with Pulp responses.

examples/search-repo-content
```diff
@@ -0,0 +1,61 @@
...
+    parser.add_argument("--insecure", default=False, action="store_true")
+
+    p = parser.parse_args()
+
+    if p.debug:
+        logging.getLogger("pubtools.pulplib").setLevel(logging.DEBUG)
+        log.setLevel(logging.DEBUG)
+
+    client = make_client(p)
+    units = client.search_repository_content(p.repo_id, type_ids=p.content_type)
```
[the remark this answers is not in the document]
nathanegillett: Well, I don't think it makes sense to repeat the same operation but I can change this to use Repository's method if that's more relevant.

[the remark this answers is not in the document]
nathanegillett: Or maybe I'll add a commented section showing how one might use Repository's method as an alternative.

pubtools/pulplib/_impl/model/unit/modulemd_defaults.py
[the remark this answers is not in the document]
JayZ12138: hmm, if I do `grep repository_id . -r` under pubtools/_impl/, it shows me only one usage of repository_id, which is used as a variable in client.py. Why do you think it's pushing repository_id instead of repo_id?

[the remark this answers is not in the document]
nathanegillett: Huh. Okay, yeah. Maintenance and distributor models use "repo_id" while get_repository uses "repository_id". There are also some private methods in client.py using "repo_id", but that's less of a concern.

I really don't know which should be preferred here. I seem to remember @rohanpm mentioning somewhere that he'd like to not use shortened words in these open-source projects, as it could confuse external contributors.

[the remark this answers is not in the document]
JayZ12138: Yup, you're right. In Rohan's knowledge share about API design, he suggested not using abbreviated terms, such as repository_id vs repo_id. But it's an exception here, since repo_id has been used before and consistency should be considered first to avoid confusion.

nathanegillett:
@jbruzl, I think you're the only code owner for this repo still around.

PR #73, opened 2019-12-04
Fix export an empty maintenance would crash

JayZ12138:
Now if the report is empty and last_updated field is None, then while
exporting report, current datetime will be used to aviod crashing.

### DIFF
--- a/pubtools/pulplib/_impl/model/convert.py
+++ b/pubtools/pulplib/_impl/model/convert.py
@@ -32,4 +32,7 @@ def read_timestamp(value):
 
 
 def write_timestamp(value):
+    # value could be None if the report is empty
+    if value is None:
+        value = datetime.datetime.utcnow()
     return value.strftime("%Y-%m-%dT%H:%M:%SZ")
--- a/pubtools/pulplib/_impl/model/convert.py
+++ b/pubtools/pulplib/_impl/model/convert.py
@@ -32,7 +32,7 @@ def read_timestamp(value):
 
 
 def write_timestamp(value):
-    # value could be None if the report is empty
+    # defaults to current time if value is None
     if value is None:
         value = datetime.datetime.utcnow()
     return value.strftime("%Y-%m-%dT%H:%M:%SZ")
--- a/tests/maintenance/test_maintenance.py
+++ b/tests/maintenance/test_maintenance.py
@@ -50,7 +50,7 @@ def test_load_invalid_report_raise_exception():
 
 def test_create_export_report():
     data = {
-        "last_updated": "2019-08-15T14:21:12Z",  # invalid timestamp
+        "last_updated": "2019-08-15T14:21:12Z",
         "last_updated_by": "pubtools.pulplib",
         "repos": {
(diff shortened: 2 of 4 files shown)


tests/maintenance/test_maintenance.py
```diff
@@ -92,3 +92,12 @@ def test_report_add_remove():
...
     assert report.last_updated_by == "jazhang"
+
+
+def test_export_empty_report():
+    # create an empty report with out data passed
+    report = MaintenanceReport()
+    assert report.last_updated is None
+
+    # export empty report shouldn't have problem
+    report._export_dict()
```
rajulkumar: may validate the last_updated as current date as well

pubtools/pulplib/_impl/model/convert.py
```diff
@@ -32,4 +32,7 @@ def read_timestamp(value):
 
 
 def write_timestamp(value):
+    # value could be None if the report is empty
```
rajulkumar: I think it will be better to keep this method generic and move setting the value to current date in _export_dict()

JayZ12138: Hm, I did it here since the MaintenanceEntry object could also have the same problem, which has a timestamp, defaulted to None as well.

tests/maintenance/test_maintenance.py
JayZ12138: I didn't remember how the timestamp was tested..., in this case, what I can think of is mock the datetime.datetime.utcnow() with a fake timestamp and check it, but doesn't make much sense to me. Do you think this is good or just a waste of lines? any other test methods?

JayZ12138: I guess I can check the output timestamp, see if it's a timestamp with right format.

pubtools/pulplib/_impl/model/convert.py
rajulkumar: That sounds fair. Since this is a generic method and not tied to maintenance_report only, please modify this comment to something like "defaults to current time if value is None"

tests/maintenance/test_maintenance.py
rajulkumar: Checking the timestamp to be the current time/mocked time in case of an empty report would have been better. But looking at the bigger picture, adding the timestamp is to ensure the report doesn't crash. So checking if timestamp exist should be fine too.

rajulkumar:
```
current_date = datetime.datetime.now()
    with patch.object(datetime, "datetime", Mock(wraps=datetime.datetime)) as mocked_datetime:
        mocked_datetime.now.return_value = current_date
        report = MaintenanceReport()._export_dict()

    assert report.get('last_updated') == datetime.datetime.strftime(current_date, "%Y-%m-%dT%H:%M:%SZ")
```

pubtools/pulplib/_impl/model/convert.py
JayZ12138: makes sense.

PR #74, opened 2019-12-11
give last_updated_by a valid default value if it's None when export

JayZ12138:
If a user calls set-maintenance but doesn't do anything and the report
was originally empty, then the last_updated_by field would be None,
this would lead to crash when set-maintenance is called again, since
the last_updated_by field couldn't be None.
Now, while exporting the report, make it to default value if the field
last_updated_by is None.

### DIFF
--- a/pubtools/pulplib/_impl/model/__init__.py
+++ b/pubtools/pulplib/_impl/model/__init__.py
@@ -1,8 +1,4 @@
-from .common import (
-    PulpObject,
-    DetachedException,
-    InvalidDataException,
-)
+from .common import PulpObject, DetachedException, InvalidDataException
 from .repository import (
     Repository,
     YumRepository,
--- a/pubtools/pulplib/_impl/model/__init__.py
+++ b/pubtools/pulplib/_impl/model/__init__.py
@@ -1,4 +1,8 @@
-from .common import PulpObject, DetachedException, InvalidDataException
+from .common import (
+    PulpObject,
+    DetachedException,
+    InvalidDataException,
+)
 from .repository import (
     Repository,
     YumRepository,
--- a/pubtools/pulplib/_impl/model/maintenance.py
+++ b/pubtools/pulplib/_impl/model/maintenance.py
@@ -219,9 +219,4 @@ class MaintenanceReport(object):
             if entry.repo_id not in repo_ids:
                 new_entries.append(entry)
 
-        return attr.evolve(
-            self,
-            entries=new_entries,
-            last_updated_by=owner,
-            last_updated=datetime.datetime.utcnow(),
-        )
+        return attr.evolve(self, last_updated_by=owner, entries=new_entries)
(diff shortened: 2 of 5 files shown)


pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -4,11 +4,7 @@
 from attr import validators
 from more_executors.futures import f_proxy
 
-from ..common import (
```
midnightercz: unrelated change

pubtools/pulplib/_impl/model/__init__.py
```diff
@@ -1,8 +1,4 @@
-from .common import (
```
midnightercz: unrelated change

JayZ12138: It's done by black automatically, not sure why.

JayZ12138:
> It's little strange that you don't have to change test_create_export_report, could you check that?

Since the data given in the test has the last_updated_by value.

pubtools/pulplib/_impl/model/__init__.py
midnightercz: can you then fix it please?

JayZ12138: sure, let me try.

JayZ12138:
> I'd also prefer seeing the style-only changes excluded from the commit. If they're introduced because black behavior changed, I think a commit/PR dedicated to just fixing the style is the way to go.

Seems like it's because I'm using an older version of black. 
I was using black 19.3 while the newest version is 19.10. With the old version, it reformats that line automatically, after I upgrade it to the newest version, either format is acceptable by black, so I can change it back.

JayZ12138:
> I just realized that no autotests had to be updated for these behavior changes. Seems like we are missing some tests we really ought to have, then. Are you willing to look into that as part of this PR?

sure.

pubtools/pulplib/_impl/model/maintenance.py
```diff
@@ -219,4 +219,9 @@ def remove(self, repo_ids, **kwargs):
             if entry.repo_id not in repo_ids:
                 new_entries.append(entry)
 
-        return attr.evolve(self, last_updated_by=owner, entries=new_entries)
+        return attr.evolve(
+            self,
+            entries=new_entries,
+            last_updated_by=owner,
+            last_updated=datetime.datetime.utcnow(),
```
rohanpm: Could you explain this part a little?  The change here was to add/overwrite last_updated on any change.  It might be correct, but all of the commit message, PR message and changelog only refer to changes to last_updated_by, a different field.

JayZ12138: Hm, I thought the problem was here for the first time and did this change, but it's not actually. 
I'll remove this since it's unrelated change

PR #78, opened 2020-02-12
Support pulp sync repository operation

midnightercz:
Repository.sync() operation is now supported for Yum,File and
ContainerImage Repository


pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -14,6 +14,21 @@
...
+class FileSyncOptions(SyncOptions):
+    """Options controlling a file repository
+    :meth:`~pubtools.pulplib.FileRepository.sync`.
+    """
+
+    remove_missing = pulp_attrib(default=False, type=bool)
+    """If true, as the repository is synchronized, old rpms will be removed.
+    """
+
+    validate = pulp_attrib(default=None, type=bool)
```
rohanpm: This seems like another good candidate for omitting.  If you set validate=False, then file corruption is silently ignored... seems odd, why would anyone want to do that?  Can we just force it as True always?

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -1,9 +1,86 @@
...
+class YumSyncOptions(SyncOptions):
+    """Options controlling a container repository
+    :meth:`~pubtools.pulplib.YumRepository.sync`.
+    """
+
+    query_auth_token = pulp_attrib(default=None, type=str)
+    """An authorization token that will be added to every request made to the feed URL's server
+    """
+
+    validate = pulp_attrib(default=None, type=bool)
```
rohanpm: Same comment as with the FileSyncOptions, can we leave it out please?  Setting validate=False seems useless.

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -1,9 +1,86 @@
...
+    Defaults to source checksum type of sha256
+    """
+
+    num_retries = pulp_attrib(default=None, type=int)
+    """Number of times to retry before declaring an error during repository synchronization
+
+    Default is to 2.
+    """
+
+    copy_children = pulp_attrib(default=None, type=bool)
```
rohanpm: Does this actually do anything / make sense here?

I looked at the docs at https://docs.pulpproject.org/en/2.14/plugins/pulp_rpm/tech-reference/yum-plugins.html

It says "Supported only as an override config option to a repository copy command", so I suppose the option is useless when doing a sync?

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -1,9 +1,86 @@
...
+
+    require_signature = pulp_attrib(default=None, type=bool)
+    """Requires that imported packages like RPM/DRPM/SRPM should be signed
+    """
+
+    allowed_keys = pulp_attrib(default=None, type=str)
+    """Comma-separated list of allowed signature key IDs that imported packages can be signed with
+    """
+
+    recursive = pulp_attrib(default=None, type=bool)
```
rohanpm: I don't think this one makes sense either.  I am pretty sure recursive applies during copy only, not sync.  In general can we please only include options here if we understand the use-case for them?

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -1,9 +1,86 @@
...
+    """If true, units are copied together with the latest versions of their dependencies
+    """
+
+    recursive_conservative = pulp_attrib(default=None, type=bool)
```
rohanpm: Same comment as for `recursive`.

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -1,9 +1,86 @@
...
+    """If true, units are copied together with their dependencies, unless those are already satisfied by the content in the target repository.
+    """
+
+    additional_repos = pulp_attrib(default=None, type=bool)
```
rohanpm: Same comment as for `additional_repos`.

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -1,9 +1,86 @@
...
+
+    force_full = pulp_attrib(default=None, type=bool)
+    """Boolean flag. If true, full re-sync is triggered.
+    """
+
+    require_signature = pulp_attrib(default=None, type=bool)
+    """Requires that imported packages like RPM/DRPM/SRPM should be signed
+    """
+
+    allowed_keys = pulp_attrib(default=None, type=str)
```
rohanpm: I think this is the wrong type.  I know the Pulp documentation says it is a comma-separated string, but it's wrong, see https://pulp.plan.io/issues/2569.  Seems this should be a list of string.

tests/fake/test_fake_sync.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # The future should resolve successfully
+    tasks = sync_f.result()
+
+    # It should have returned at least one successful task.
+    assert tasks
+    for task in tasks:
+        assert task.succeeded
+
+    # The change should be reflected in the controller's publish history
```
rohanpm:
```suggestion
    # The change should be reflected in the controller's sync history
```

tests/fake/test_fake_sync.py
```diff
@@ -0,0 +1,49 @@
...
+    history = controller.sync_history
+
+    assert len(history) == 1
+    assert history[0].repository.id == "repo1"
+    assert history[0].tasks == tasks
+
+
+def test_publish_absent_raises():
```
rohanpm:
```suggestion
def test_sync_absent_raises():
```

tests/fake/test_fake_sync.py
```diff
@@ -0,0 +1,49 @@
...
+    """repo.publish() of a nonexistent repo raises."""
```
rohanpm:
```suggestion
    """repo.sync() of a nonexistent repo raises."""
```

tests/fake/test_fake_sync.py
```diff
@@ -0,0 +1,49 @@
...
+    controller.insert_repository(YumRepository(id="repo1"))
+
+    client = controller.client
+    repo_copy1 = client.get_repository("repo1")
+    repo_copy2 = client.get_repository("repo1")
+
+    # If I delete the repo through one handle...
+    assert repo_copy1.delete().result()
+
+    # ...then publish through the other handle becomes impossible
```
rohanpm:
```suggestion
    # ...then sync through the other handle becomes impossible
```

tests/repository/test_sync.py
```diff
@@ -0,0 +1,75 @@
...
+@pytest.fixture
+def fixture_search_task_response(requests_mocker):
+    requests_mocker.post(
+        "https://pulp.example.com/pulp/api/v2/tasks/search/",
+        [{"json": [{"task_id": "task1", "state": "finished"}]}],
+    )
+
+
+def test_detached():
+    """publish raises if called on a detached repo"""
```
rohanpm:
```suggestion
    """sync raises if called on a detached repo"""
```

tests/repository/test_sync.py
```diff
@@ -0,0 +1,75 @@
...
+    with pytest.raises(DetachedException):
+        Repository(id="some-repo").sync()
+
+
+def test_sync(
```
rohanpm: The test name and doc string doesn't match at all what's happening in the test body... should this be named something like `test_sync_missing_feed`  with doc `sync fails if no feed is provided` ?

tests/repository/test_sync.py
```diff
@@ -0,0 +1,75 @@
...
+        repo.sync().result()
+        assert "Exception should have been raised"
+    except ValueError:
+        pass
+
+
+def test_sync_with_options(
+    requests_mocker, client, fixture_sync_async_response, fixture_search_task_response
+):
+    """publish passes expected config into distributors based on publish options"""
```
rohanpm: Doc string is incorrect.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -565,3 +565,18 @@ def map_404_to_none(exception):
         )
 
         return f_map(response, error_fn=map_404_to_none)
+
+    def _do_sync(self, repo_id, sync_options):
+        if not sync_options["feed"]:
```
rohanpm: Should we have this check actually?

1. Like any other options, it's possible for feed to be stored on the importer config itself, in which case it's not necessary to pass it.  Ideally the library doesn't block this scenario.
2. If the Pulp server really needs a feed (because it's missing from importer config), it should already be returning a meaningful error message, so we shouldn't have to duplicate the same error check here.

pubtools/pulplib/_impl/fake/controller.py
```diff
@@ -112,6 +112,21 @@ def publish_history(self):
...
 
+    @property
+    def sync_history(self):
+        """A list of repository syncs triggered via this client.
+
+        Each element of this list is a named tuple with the following attributes,
+        in order:
+
+            ``repository``:
+                :class:`~pubtools.pulplib.Repository` for which publish was triggered
```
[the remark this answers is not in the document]
midnightercz: yeah, good idea.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -62,6 +58,53 @@ class PublishOptions(object):
...
+    ssl_client_cert = pulp_attrib(default=None, type=str)
+    """Certificate used as the client certificate when synchronizing the repository
+    """
+
+    ssl_client_key = pulp_attrib(default=None, type=str)
+    """Private key to the certificate specified in ssl_client_cert
+    """
+
+    max_speed = pulp_attrib(default=None, type=int)
+    """Representing the maximum speed that the importer should be allowed to transfer
```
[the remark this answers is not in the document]
midnightercz: I copy pasted this from pulp docs. I newer version they have description that makes more sense. I will update this.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -343,6 +386,34 @@ def publish(self, options=PublishOptions()):
...
+        Raises:
+            DetachedException
+                If this instance is not attached to a Pulp client.
+        """
+        if not self._client:
+            raise DetachedException()
+
+        return f_proxy(
+            self._client._do_sync(
+                self.id, asdict(options, filter=lambda name, val: val is not None)
```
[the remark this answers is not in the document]
midnightercz: in all _do method I observed (and maybe I'm wrong) that all arguments are basic types (str, int, dict,...) so I wanted to stick with that.

pubtools/pulplib/_impl/model/repository/container.py
```diff
@@ -1,8 +1,35 @@
...
 
 
+@attr.s(kw_only=True, frozen=True)
+class ContainerSyncOptions(SyncOptions):
+    """Options controlling a container repository
+    :meth:`~pubtools.pulplib.ContainerImageRepository.sync`.
+    """
+
+    upstream_name = pulp_attrib(default=None, type=str)
+    """The name of the repository to import from the upstream repository.
```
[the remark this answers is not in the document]
midnightercz: sure, no problem. I only copied what is in pulp docs.

pubtools/pulplib/_impl/model/repository/container.py
```diff
@@ -1,8 +1,35 @@
...
+    """
+
+    tags = pulp_attrib(default=None, type=str)
```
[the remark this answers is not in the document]
midnightercz: right, I need to fix this

pubtools/pulplib/_impl/model/repository/yum.py
midnightercz: yeah, thanks for catching this one.

pubtools/pulplib/_impl/client/client.py
midnightercz: Well, I wanted to do this "user-friendly" and fails asap if there's obvious error. As you said feed can be used from importer but to do that I would have to implement the model and another support code, which is something I'm not that much willing to do now due lots of changes needed in another projects right now. So I thought we still override feed everytime so let's check that feed is not empty here and in the future we can implement importer model in repository model.
If workflow is: get repository model, run sync, then at that time you will have already information about importer configuration and you can decide if there's feed or not and therefore if "external" feed is required. 
Another reason is log message few lines bellow this one which tells you: Syncing repository with feed: xxx.
But maybe it's unnecessary complicated and I should rather let pulp return error here.

tests/repository/test_sync.py
midnightercz: yeah, makes sense

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -62,6 +58,53 @@ class PublishOptions(object):
     """
 
 
+@attr.s(kw_only=True, frozen=True)
+class SyncOptions(object):
+    """Options controlling a repository
+    :meth:`~pubtools.pulplib.Repository.sync`.
+    """
+
+    feed = pulp_attrib(type=str)
```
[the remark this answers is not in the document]
midnightercz: yeah, importer is problem here. In theory you can sync without the feed and default feed is then used from importer config.But there's no implementation of importer model yet.

midnightercz:
> Most of my comments are about the SyncOptions classes. I think that you've added an option for pretty much everything which exists in Pulp, but not all of them make sense.

I agree we don't use them in our toolchain - yet. This is I think third time I decided to use pubtools-pulplib for implementation of some new feature in our toolchain and I found out that there's missing functionality I need in this library. So I thought it would be better to rather implement everything that is supported then in the future find out you need to add something again.

pubtools/pulplib/_impl/client/client.py
rohanpm: I didn't mean that this library needs to fetch the importer and use it to initialize feed - I meant that you can just do a sync request to Pulp without including any feed, and it'll work if the importer already has a feed or it'll return a meaningful error if it doesn't.  So you wouldn't have to add Importer models yet.

In any case though, I don't consider this a blocker for the merge request, seeing as it can be easily changed later without breaking compatibility.

pubtools/pulplib/_impl/fake/controller.py
```diff
@@ -112,6 +112,21 @@ def publish_history(self):
...
+            ``tasks``:
+                list of :class:`~pubtools.pulplib.Task` generated as a result
+                of this publish
+        """
```
rohanpm:
```suggestion
            ``sync_config``:
                :class:`~pubtools.pulplib.SyncConfig` (of the appropriate subclass) used for this sync
        """
```

pubtools/pulplib/_impl/fake/controller.py
rohanpm:
```suggestion
                :class:`~pubtools.pulplib.Repository` for which sync was triggered
```

pubtools/pulplib/_impl/fake/controller.py
```diff
@@ -112,6 +112,21 @@ def publish_history(self):
...
+        """A list of repository syncs triggered via this client.
+
+        Each element of this list is a named tuple with the following attributes,
+        in order:
+
+            ``repository``:
+                :class:`~pubtools.pulplib.Repository` for which publish was triggered
+            ``tasks``:
+                list of :class:`~pubtools.pulplib.Task` generated as a result
+                of this publish
```
rohanpm:
```suggestion
                of this sync
```

pubtools/pulplib/__init__.py
```diff
@@ -16,6 +16,9 @@
     ModulemdDefaultsUnit,
     Distributor,
     PublishOptions,
+    FileSyncOptions,
+    ContainerSyncOptions,
+    YumSyncOptions,
```
rohanpm: Can we list SyncOptions too please?  The reason is mainly for docs, we can't properly refer to the class in docs if it's not exported somewhere.

```suggestion
    YumSyncOptions,
    SyncOptions,
```

pubtools/pulplib/_impl/model/repository/base.py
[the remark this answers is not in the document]
rohanpm: It makes sense to drop this suggestion if you'll keep that "feed" check you have elsewhere.

PR #79, opened 2020-02-19
Minor doc improvements for recent changes

rohanpm:
- Add missing CHANGELOG entry for new API

- Fix mispelled ModulemdDefaultsUnit class in docs

- Add all the SyncOptions classes into docs

- Fix wrong copy-pasted summary on YumSyncOptions

- Add "versionadded" a few places for sync to help record which
  minimum version is needed to use this feature

- Add a "seealso" on SyncOptions to help find the other related
  classes

- Fix links from SyncOptions back to sync method (linking to
  inherited sync() method on repository subclasses doesn't seem to
  work)

- Tweak how the default argument to sync() is set; it's functionally
  equivalent to how it was before, but renders more nicely in docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- N/A
+### Added
+- Introduced `Repository.sync` API (and associated `SyncOptions` classes)
+  for synchronizing Pulp repositories.
 
 ## [2.4.0] - 2020-01-13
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- N/A
+### Added
```
midnightercz: can you move this to [2.5.0]  so we can release the changes after PR is merged?

rohanpm: That's normally done right at the time of release because you don't know the date of the release otherwise.

PR #84, opened 2020-03-26
Added population attributes for yum repositories

midnightercz:
Added 'population_sources' and 'ubi_population'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - sourcerpm attribute for Rpm unit
+- Added 'population_sources' and 'ubi_population'
 
 ## [2.5.0] - 2020-02-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - sourcerpm attribute for Rpm unit
-- Added 'population_sources' and 'ubi_population'
+###Added
+- Introduced 'population_sources' and 'ubi_population' attributes for yum repository
 
 ## [2.5.0] - 2020-02-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - sourcerpm attribute for Rpm unit
-###Added
 - Introduced 'population_sources' and 'ubi_population' attributes for yum repository
 
 ## [2.5.0] - 2020-02-25
--- a/pubtools/pulplib/_impl/model/repository/yum.py
+++ b/pubtools/pulplib/_impl/model/repository/yum.py
@@ -70,6 +70,18 @@ class YumRepository(Repository):
 
     type = pulp_attrib(default="rpm-repo", type=str, pulp_field="notes._repo-type")
 
+    population_sources = pulp_attrib(
+        default=(), type=tuple, converter=tuple, pulp_field="notes.population_sources"
+    )
+    """List of repositories used for populate the repository
+    """
+
+    ubi_population = pulp_attrib(
+        default=False, type=bool, pulp_field="notes.ubi_population"
+    )
+    """Flag indicating whether repo should be populated or not
+    """
+
     mutable_urls = attr.ib(
         default=attr.Factory(lambda: FrozenList(["repodata/repomd.xml"])),
         type=list,
--- a/pubtools/pulplib/_impl/model/repository/yum.py
+++ b/pubtools/pulplib/_impl/model/repository/yum.py
@@ -73,7 +73,7 @@ class YumRepository(Repository):
     population_sources = pulp_attrib(
         default=(), type=tuple, converter=tuple, pulp_field="notes.population_sources"
     )
-    """List of repositories used for populate the repository
+    """List of repository IDs used to populate this repository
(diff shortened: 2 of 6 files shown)


pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -70,6 +70,18 @@ class YumRepository(Repository):
...
+        default=(), type=tuple, converter=tuple, pulp_field="notes.population_sources"
+    )
+    """List of repositories used for populate the repository
+    """
+
+    ubi_population = pulp_attrib(
+        default=False, type=bool, pulp_field="notes.ubi_population"
+    )
+    """Flag indicating whether repo should be populated or not
+    """
```
rbikar: This change isn't mentioned in the commit msg. Please update it, thx.

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -70,6 +70,18 @@ class YumRepository(Repository):
 
     type = pulp_attrib(default="rpm-repo", type=str, pulp_field="notes._repo-type")
 
+    population_sources = pulp_attrib(
+        default=(), type=tuple, converter=tuple, pulp_field="notes.population_sources"
```
rohanpm: Can you please make this consistent with the other "immutable collection" attributes, i.e. use FrozenList?

So that would be something like:

```suggestion
        default=attr.Factory(FrozenList), type=list, converter=FrozenList, pulp_field="notes.population_sources"
```

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -70,6 +70,18 @@ class YumRepository(Repository):
...
+    )
+    """List of repositories used for populate the repository
```
rohanpm:
```suggestion
    """List of repository IDs used to populate this repository
```

It is repo IDs, right?  Not actually Repository objects.

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -70,6 +70,18 @@ class YumRepository(Repository):
...
+    """
+
+    ubi_population = pulp_attrib(
+        default=False, type=bool, pulp_field="notes.ubi_population"
+    )
+    """Flag indicating whether repo should be populated or not
```
rohanpm: population_sources gives the impression of something a bit generic but this one is named specifically relating to UBI.  I guess the doc should be slightly more detailed then, like:

```suggestion
    """Flag indicating whether repo should be populated or not for the purposes of UBI
```

tests/repository/test_yum_repository.py
```diff
@@ -54,5 +54,21 @@ def test_from_data_skip_rsync_repodata():
             ],
         }
     )
-
     assert repo.skip_rsync_repodata
+
+
+def test_populate_attrs():
+    """skip_rsync_repodata is initialized from distributors when possible"""
```
rohanpm: This doc string is wrong, copy-pasted I guess?

pubtools/pulplib/_impl/schema/repository.yaml
```diff
@@ -83,6 +83,15 @@ properties:
       pub_temp_repo:
         type: boolean
 
+      # True if repository supposed to be populated from repositories
+      # mentioned in pupulation_sources
```
rohanpm:
```suggestion
      # mentioned in population_sources
```

pubtools/pulplib/_impl/schema/repository.yaml
```diff
@@ -83,6 +83,15 @@ properties:
...
+      ubi_population:
+        type: boolean
+
+      # List of repositories used for populating the repository
+      population_sources:
+        type: array
```
rohanpm: Can we please validate the contents of the array as well, it's easy:

```suggestion
        type: array
        items:
          type: string
```

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- n/a
+- Added 'population_sources' and 'ubi_population'
```
rajulkumar, line 10: This could be a little more explicit like "Added 'population_sources' list and 'ubi_population' flag as attributes to yum repository'. May also add '###Added' heading here.

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -70,6 +70,21 @@ class YumRepository(Repository):
...
+        converter=FrozenList,
+        pulp_field="notes.population_sources",
+    )
+    """List of repository IDs used to populate this repository
+    """
+
+    ubi_population = pulp_attrib(
+        default=False, type=bool, pulp_field="notes.ubi_population"
+    )
+    """Flag indicating whether repo should be populated or not for the purposes of UBI
```
rajulkumar, line 85: may mention '...populated from population_sources' just to be consistent with the comment in repository.yaml

CHANGELOG.md
nathanegillett: +1

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - sourcerpm attribute for Rpm unit
+###Added
```
rohanpm, line 12: 1. Added header need not be duplicated.
```suggestion
```

2. It's not valid markdown, there would have to be a space between ### and Added.

rohanpm: Do you want project owners to fix up minor issues like this on your behalf at merge time?

PR #86, opened 2020-04-07
Added client.search_content

midnightercz:
New client method search_content search content by given
 criteria in pulp regardless of repository.
All unit models were also extended with repository_memberships
attribute indicating what repositories given unit belongs to.

### DIFF
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -264,7 +264,9 @@ class Client(object):
         criteria=None,
     ):  # pylint:disable = too-many-arguments
 
-        if not isinstance(resource_types, list) or isinstance(resource_types, tuple):
+        if not isinstance(resource_types, list) and not isinstance(
+            resource_types, tuple
+        ):
             resource_types = [resource_types]
 
         responses = []
@@ -275,7 +277,7 @@ class Client(object):
                 self._url, "pulp/api/v2/%s/%s/" % (resource_type, search_type)
             )
             urls.append(url)
-            prepared_search = search_for_criteria(criteria, return_type, type_ids_accum)
+            prepared_search = search_for_criteria(criteria, return_type)
 
             search = {
                 "criteria": {
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -14,7 +14,7 @@ from more_executors.futures import f_map, f_flat_map, f_return, f_proxy
 from ..page import Page
 from ..criteria import Criteria
 from ..model import Repository, MaintenanceReport, Distributor, Unit
-from .search import search_for_criteria
+from .search import search_for_criteria, TypeIdAccumulator
 from .errors import PulpException
 from .poller import TaskPoller
 from . import retry
@@ -203,6 +203,33 @@ class Client(object):
             Repository, "repositories", criteria=criteria, search_options=search_options
         )
 
+    def search_units_by_type(self, content_type, criteria=None):
+        """Search for units of single type across all repositories.
+
+        Args:
+            criteria (:class:`~pubtools.pulplib.Criteria`)
+                A criteria object used for this search.
+                If None, search for all repositories.
+
+        Returns:
+            Future[:class:`~pubtools.pulplib.Page`]
+                A future representing the first page of results.
+
+                Each page will contain a collection of
+                :class:`~pubtools.pulplib.Repository` objects.
+        """
+        type_ids_acc = TypeIdAccumulator()
+        type_ids_acc.no_accumulate_error = (
+            "Cannot query for _content_type_id when search over all repositories"
+        )
+        with type_ids_acc.no_accumulate:
+            return self._search(
+                Unit,
+                "content/units/%s" % content_type,
+                criteria=criteria,
+                type_ids_accum=type_ids_acc,
+            )
+
     def search_distributor(self, criteria=None):
         """Search the distributors matching the given criteria.
 
@@ -229,11 +256,12 @@ class Client(object):
         search_type="search",
         search_options=None,
         criteria=None,
+        type_ids_accum=None,
     ):  # pylint:disable = too-many-arguments
         url = os.path.join(
             self._url, "pulp/api/v2/%s/%s/" % (resource_type, search_type)
         )
-        prepared_search = search_for_criteria(criteria, return_type)
+        prepared_search = search_for_criteria(criteria, return_type, type_ids_accum)
 
         search = {
             "criteria": {
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -9,12 +9,12 @@ from six.moves import StringIO
 
 import requests
 from more_executors import Executors
-from more_executors.futures import f_map, f_flat_map, f_return, f_proxy
+from more_executors.futures import f_map, f_flat_map, f_return, f_proxy, f_sequence
 
 from ..page import Page
 from ..criteria import Criteria
 from ..model import Repository, MaintenanceReport, Distributor, Unit
-from .search import search_for_criteria, TypeIdAccumulator
+from .search import search_for_criteria
 from .errors import PulpException
 from .poller import TaskPoller
 from . import retry
@@ -203,7 +203,7 @@ class Client(object):
             Repository, "repositories", criteria=criteria, search_options=search_options
         )
 
-    def search_content_by_type(self, content_type, criteria=None):
+    def search_content(self, criteria=None):
         """Search for units of single type across all repositories.
 
         Args:
@@ -220,17 +220,21 @@ class Client(object):
 
         .. versionadded:: 2.6.0
         """
-        type_ids_acc = TypeIdAccumulator()
-        type_ids_acc.no_accumulate_error = (
-            "Cannot query for _content_type_id when search over all repositories"
-        )
-        with type_ids_acc.no_accumulate:
-            return self._search(
-                Unit,
-                "content/units/%s" % content_type,
-                criteria=criteria,
-                type_ids_accum=type_ids_acc,
+
+        server_type_ids = self.get_content_type_ids()
+        prepared_search = search_for_criteria(criteria, Unit, None)
+        type_ids = prepared_search.type_ids
+        if not type_ids:
+            type_ids = server_type_ids
+        missing = set(type_ids) - set(server_type_ids)
+        if missing:
+            raise ValueError(
+                "Content type: %s is not supported by server" % ",".join(missing)
             )
+
+        return self._search(
+            Unit, ["content/units/%s" % x for x in type_ids], criteria=criteria
+        )
 
     def search_distributor(self, criteria=None):
         """Search the distributors matching the given criteria.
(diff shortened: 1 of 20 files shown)


pubtools/pulplib/_impl/client/client.py
```diff
@@ -203,6 +203,33 @@ def search_repository(self, criteria=None):
             Repository, "repositories", criteria=criteria, search_options=search_options
         )
 
+    def search_units_by_type(self, content_type, criteria=None):
+        """Search for units of single type across all repositories.
+
+        Args:
+            criteria (:class:`~pubtools.pulplib.Criteria`)
+                A criteria object used for this search.
+                If None, search for all repositories.
```
rohanpm, line 212: This doc string doesn't seem right.  Copy-pasted?  Aren't we *always* searching over all repositories, and this criteria is applied on units, not repositories?

pubtools/pulplib/_impl/client/client.py
```diff
@@ -203,6 +203,33 @@ def search_repository(self, criteria=None):
...
+
+        Returns:
+            Future[:class:`~pubtools.pulplib.Page`]
+                A future representing the first page of results.
+
+                Each page will contain a collection of
+                :class:`~pubtools.pulplib.Repository` objects.
```
rohanpm, line 219: This is false, right?

pubtools/pulplib/_impl/client/client.py
```diff
@@ -203,6 +203,33 @@ def search_repository(self, criteria=None):
             Repository, "repositories", criteria=criteria, search_options=search_options
         )
 
+    def search_units_by_type(self, content_type, criteria=None):
```
rohanpm, line 206: This API doesn't seem to make sense when taken together with the existing Repository.search_content.

- Why is Repository.search_content named search_content but this is named search_units* ?
- Why does this API require the caller to pass content_type as a separate argument while Repository.search_content instead allows putting content_type_id into Criteria?  

To elaborate on the second point:

```
# If I have a repo, I can find all units of type "foo" matching certain criteria like this:
repo.search_content(Criteria.and_(Criteria.with_field("content_type_id", "foo"), <other_criteria>))

# But if I want to search across *all* repos I instead need to do:
client.search_units_by_type("foo", <other_criteria>)
```

Why the inconsistency?  There doesn't seem to be any reason why per-repo search and all-repo search methods need to have different signatures.

pubtools/pulplib/_impl/model/unit/file.py
```diff
@@ -31,6 +32,17 @@ class FileUnit(Unit):
         default="iso", type=str, pulp_field="_content_type_id"
     )
 
+    repository_memberships = pulp_attrib(
+        default=attr.Factory(FrozenList),
+        type=list,
+        converter=FrozenList,
+        pulp_field="repository_memberships",
+    )
+    """List of repositories which unit belongs to
```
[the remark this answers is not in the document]
midnightercz: When I added this to unit base class attr lib complained about mandatory attributes were defined after optional one is defined. It was just a warning but I wanted to avoid that.

pubtools/pulplib/_impl/client/client.py
midnightercz: What if you pass there criteria like:
Criteria.with_field("content_type_id", Matcher._in(["rpm","srpm"]))
Which content_type_id should be then used for the search?
That's why I rather made it inconsistent.

rohanpm: It should then do more than one request for the search.  It's not much different from pagination - it's already undefined how many requests are done to the server for any particular search.

midnightercz: ok, but what if user specify criteria like 
`
Criteria.with_field("content_type_id", Matcher._regex("?rpm"))
`
It's most probably not supported right now, but it could be in the future if TypeIdsAccumulator.accumulate_from_match changes.
That would basically mean, I need to gather available content types from server and do the matching locally. And that I wanted to avoid.
Reasonable could be subsclassing TypeIdsAccumulator with restriction only for matchers like $eq and $in.
I would like to hear your preferred solution for this before I got o start changing anything.

pubtools/pulplib/_impl/schema/unit.yaml
```diff
@@ -123,6 +138,7 @@ definitions:
     - context
     - arch
 
+
```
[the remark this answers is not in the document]
midnightercz: yep

pubtools/pulplib/_impl/client/client.py
```diff
@@ -203,6 +203,35 @@ def search_repository(self, criteria=None):
             Repository, "repositories", criteria=criteria, search_options=search_options
         )
 
+    def search_content_by_type(self, content_type, criteria=None):
```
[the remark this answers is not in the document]
midnightercz: well, all this is about "user" comfort we you want to provide here. Of course use case when user wants to get both rpm and srpm in one method call makes sense. Due to pulp implementation details it's not possible to do such call in one request, therefore logic for getting more then one content type is simply do more then one request and merge results together. So you can implement search_content_by_type to do that for you or you can leave it up to user.
First approach provides more user comfort, however require more code to handle possible user inputs (see my conversation with Rohan above), second approach makes code easier and user had to do those steps "manually".
So I decided to rather leave it up to user as sequence of manual steps needed here is very simple compared to extra code needed for proper handling user input. 

Benefit vs effort needed for first option just seemed very unbalanced to me.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -203,6 +203,35 @@ def search_repository(self, criteria=None):
...
+                A future representing the first page of results.
+
+                Each page will contain a collection of
+                :class:`~pubtools.pulplib.Unit` subclasses objects.
+
+        .. versionadded:: 2.6.0
+        """
+        type_ids_acc = TypeIdAccumulator()
+        type_ids_acc.no_accumulate_error = (
+            "Cannot query for _content_type_id when search over all repositories"
```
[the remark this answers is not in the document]
midnightercz: It says you cannot use _content_type_id in the query

tests/client/test_client.py
```diff
@@ -359,3 +360,96 @@ def test_set_maintenance(client, requests_mocker):
...
+    with pytest.raises(ValueError) as e:
+        units = client.search_content_by_type(
+            "rpm", Criteria.with_field("_content_type_id", "rpm")
+        )
+
+    assert (
+        str(e.value)
+        == "Cannot query for _content_type_id when search over all repositories"
+    )
+    # It should have issued only a single search
```
querti, line 454: This comment doesn't make much sense.

pubtools/pulplib/_impl/client/client.py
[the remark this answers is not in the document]
nathanegillett: I forgot Pulp only permits querying a single content type. I thought we had built something in to account for cases where _search_repo_units received more than one type ID but it looks like there's no handling for that and the user would get the Pulp error.

pubtools/pulplib/_impl/client/client.py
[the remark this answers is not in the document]
nathanegillett: So, with `no_accumulate`, if `accumulate_from_match` is called _at all_ it'll return a value error. I guess I'm just confused as to how we know the issue will always be with `_content_type_id` if/when that happens here. How do we know one of the existing value error messages in `accumulate_from_match` isn't more appropriate?

pubtools/pulplib/_impl/client/client.py
rohanpm: That specific example should give them a fatal error as not supported (before the request is sent to the server).  Support can be added later if there is a use-case for it.

My preferred solution is to discourage searching on content_type_id at all and let the caller instead pass the model class they're interested in, since the class already knows its associated Pulp unit types and this info ideally should not have to be duplicated by the caller.  As written in https://github.com/release-engineering/pubtools-pulplib/issues/64 :

```
unit = RpmUnit(name='bash')
crit = Criteria.for_units([unit])
# Search all bash RPMs.
# Under the hood, the library knows that RpmUnit can belong to 'rpm' or 'srpm' content types
# and so it issues search requests for both.
client.search_content(criteria=crit)
```

However I don't think this PR should be forced to implement that, a "raw" search on content_type_id is OK as long as it's consistent with repo.search_content.

pubtools/pulplib/_impl/client/client.py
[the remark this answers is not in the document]
rohanpm: > I forgot Pulp only permits querying a single content type, then thought we had built something in to account for cases where _search_repo_units received more than one type ID but it looks like there's no handling for that and the user would get the Pulp error.

No I don't think there's an error in that case.  It works, you can search for repo units over multiple type IDs at once.

There's an inconsistency in Pulp2's API here: when searching for units associated with a repo, you can search as many type IDs as you want in a single request.  When searching for units across multiple repos, you need to do a separate request per type ID.

The original approach in the pull request here has effectively passed along this inconsistency to the user of this library: if you want to search multiple type IDs in a single repo, you can do it by a single method call since Pulp works that way.  If you want to search multiple type IDs across all repos, you need one method call per type ID since Pulp works that way.

My view is that the library should instead be hiding this inconsistency rather than forcing library users to deal with it.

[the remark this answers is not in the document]
nathanegillett: Ah, I see. Thanks, @rohanpm.

tox.ini
```diff
@@ -3,7 +3,7 @@ envlist = py27,py36,static,pidiff,docs,revdep-pubtools-pulp
 
 [testenv]
 deps=-rtest-requirements.txt
-commands=pytest -v {posargs}
+commands=pytest -vv {posargs}
```
nathanegillett, line 6: This needed or was it used for local testing?

midnightercz: That was for local testing, I forgot to remove it

pubtools/pulplib/_impl/client/client.py
```diff
@@ -225,43 +258,54 @@ def search_distributor(self, criteria=None):
     def _search(
         self,
         return_type,
-        resource_type,
+        resource_types,
         search_type="search",
         search_options=None,
         criteria=None,
+        type_ids_accum=None,
```
nathanegillett, line 265: `type_ids_accum` isn't no longer used, right?

pubtools/pulplib/_impl/model/unit/file.py
```diff
@@ -31,6 +32,17 @@ class FileUnit(Unit):
         default="iso", type=str, pulp_field="_content_type_id"
     )
 
+    repository_memberships = pulp_attrib(
+        default=attr.Factory(FrozenList),
```
rohanpm, line 36: Question: this repository_memberships is only filled in for certain kinds of searches, right?  You don't get this info when you search for units via Repository.search_content?

I think then the default should rather be `None`.  If it's an empty list then there is no obvious way for a caller to tell if they got an empty repository_memberships because the unit genuinely belongs to no repository or because of how the unit was retrieved.

So, the default would be None and the docs would point out that the info is not always available.

```suggestion
        default=None,
```

pubtools/pulplib/_impl/model/unit/file.py
```diff
@@ -31,6 +32,17 @@ class FileUnit(Unit):
...
+        type=list,
+        converter=FrozenList,
+        pulp_field="repository_memberships",
+    )
+    """IDs of repositories which unit belongs to
```
rohanpm, line 41:
```suggestion
    """IDs of repositories containing the unit, or ``None`` if this information is unavailable.
```

pubtools/pulplib/_impl/fake/client.py
```diff
@@ -93,6 +93,31 @@ def search_repository(self, criteria=None):
...
+        criteria = criteria or Criteria.true()
+        out = []
+
+        # Pass the criteria through the code used by the real client to build
+        # up the Pulp query. We don't actually *use* the resulting query since
+        # we're not accessing a real Pulp server. The point is to ensure the
+        # same validation and error behavior as used by the real client also
+        # applies to the fake.
+        prepared_search = search_for_criteria(criteria, Unit)
+
```
rohanpm, line 106: The real client will raise an error somewhere around here, if the criteria included some type which is not supported by the server.

Could this client please do the same thing?  Otherwise, tests using the fake client won't be able to find bugs such as a search query including some useless wrong type id.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -225,43 +258,55 @@ def search_distributor(self, criteria=None):
...
-        prepared_search = search_for_criteria(criteria, return_type)
 
-        search = {
-            "criteria": {
-                "skip": 0,
-                "limit": self._PAGE_SIZE,
-                "filters": prepared_search.filters,
+        if not isinstance(resource_types, list) and not isinstance(
+            resource_types, tuple
+        ):
```
nathanegillett, line 269: This could be;
```suggestion
        if not isinstance(resource_types, (list, tuple)):
```

pubtools/pulplib/_impl/fake/client.py
rohanpm: I don't think the implemented logic is correct, looks like it's going to raise errors if you try to search on a content type for which there's no data.  No data doesn't mean the type ID isn't supported.

Seems like the list of types could be gathered from the classes which exist in this library, or even just set as a hardcoded list.

pubtools/pulplib/_impl/model/unit/file.py
midnightercz: yeah, this makes sense.

PR #87, opened 2020-05-01
Fix wrongly mandatory fields for modulemd_defaults

rohanpm:
If you look at the schema defined in Pulp, 'stream' and 'profiles' are
not actually mandatory (not part of the unit key).

Although I believe in practice they're expected to be defined for all
real data, the fact that they're not part of the unit key means they're
omitted from responses in some contexts (e.g. units_succcessful in
associate/unassociate calls), therefore they must not be mandatory here
or crashes will occur when handling those responses.

Declaration of 'profiles' was also tweaked for consistency with other
fields: in other classes, a value which is allowed to be None still has
a type declared (as all types are understood to be nullable), so the
fact that this field can be None is not a reason to omit a type here
either.

One test was tweaked to ensure no crashes when stream is missing.

Relates to RHELDST-861.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - sourcerpm attribute for Rpm unit
 - Introduced 'population_sources' and 'ubi_population' attributes for yum repository
 
+### Fixed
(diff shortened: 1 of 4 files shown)


rohanpm:
Did a trivial conflict resolution with the addition of repository_memberships.

PR #90, opened 2020-06-09
Use default registry_id when it's not in distributor config

emilyzheng:
For a repo which has docker_distributor_web distributor, and repo-registry-id is not set in the distributor's config, use repo id as its default registry_id.

### DIFF
--- a/pubtools/pulplib/_impl/model/repository/container.py
+++ b/pubtools/pulplib/_impl/model/repository/container.py
@@ -47,8 +47,9 @@ class ContainerImageRepository(Repository):
 
         for dist in data.get("distributors") or []:
             if dist["distributor_type_id"] == "docker_distributor_web":
-                if "repo-registry-id" in dist["config"]:
-                    out["registry_id"] = dist["config"].get("repo-registry-id")
+                registry_id = dist["config"].get("repo-registry-id")
+                if registry_id:
+                    out["registry_id"] = registry_id
                 break
 
         return out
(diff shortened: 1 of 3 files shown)


tests/repository/test_container_image_repository.py
```diff
@@ -31,3 +31,22 @@ def test_registry_id_from_distributor():
     )
 
     assert repo.registry_id == "some/repo"
+
+
+def test_default_registry_id_from_distributor():
+    """default registry_id is used when it's not defined in distributor"""
+    repo = Repository.from_data(
```
querti, line 38: Shouldn't this be ContainerImageRepository? That's where the logic is implemented.

pubtools/pulplib/_impl/model/repository/container.py
```diff
@@ -47,7 +47,8 @@ def _data_to_init_args(cls, data):
 
         for dist in data.get("distributors") or []:
             if dist["distributor_type_id"] == "docker_distributor_web":
-                out["registry_id"] = dist["config"].get("repo-registry-id")
+                if "repo-registry-id" in dist["config"]:
```
rohanpm, line 50: I checked code in pulp_docker just now and it seems that, if repo-registry-id is set to null or empty string in config, then the default behavior of using the repo id also applies.  But this logic currently doesn't match that, it only uses the repo id if repo-registry-id is *absent*.  So I think this is still not quite matching Pulp logic, and it should instead be something like:

```suggestion
                registry_id = dist["config"].get("repo-registry-id")
                if registry_id:
                    out["registry_id"] = registry_id
```

tests/repository/test_container_image_repository.py
rohanpm: This is fine, as Repository.from_data figures out the concrete subclass to use based on the data.

pubtools/pulplib/_impl/model/repository/container.py
emilyzheng: Thanks for pointing this out.

PR #92, opened 2021-03-11
Allow customizing throttle count for task executor

rbikar:
This change allows to customize count of throttle
for Client._task_executor by passing arg to Client.

In practice this will allow to run specific number of tasks
at one time.

### DIFF
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -109,6 +109,9 @@ class Client(object):
 
                 These may be used, for example, to configure the credentials
                 for the Pulp server or to use an alternative CA bundle.
+
+            task_throttle
+                If passed in kwargs, it'll be used as param to _task_executor.
         """
         self._url = url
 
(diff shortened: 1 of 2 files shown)


rbikar:
Trying to fix unrelated CI issues in https://github.com/release-engineering/pubtools-pulplib/pull/93

pubtools/pulplib/_impl/client/client.py
```diff
@@ -109,6 +109,9 @@ def __init__(self, url, **kwargs):
 
                 These may be used, for example, to configure the credentials
                 for the Pulp server or to use an alternative CA bundle.
+
+            task_throttle
+                If passed in kwargs, it'll be used as param to _task_executor.
```
rohanpm, line 114: I don't think the docs make sense as the reader of the docs is not supposed to know what a private variable like _task_executor is.

Here's a suggestion:

```suggestion
            int task_throttle
                Maximum number of queued or running tasks permitted for this client.
                If more than this number of tasks are running, the client will wait before triggering more.
                This can be used to ensure no single client overwhelms the Pulp server.
```

I'd like to suggest also moving the doc for this one above the requests.Session parameters because it's more pulp-specific while the requests.Session stuff is generic.  Docs seem to flow better than way.

rohanpm (changes requested):

rohanpm (approved):

PR #93, opened 2021-03-11
Various CI fixes

rbikar:
1) Due to problems with rust compiler, let's move to newer dist of ubuntu on travis CI.
2) Install rpm package on travis hosts
3) Fix formatting in a couple of files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 dist: bionic
 sudo: false
 language: python
+before_install:
+  # for rpm-py-installer
+  - sudo apt-get install -y rpm
 install: pip install tox
 matrix:
   include:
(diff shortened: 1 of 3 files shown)


rbikar:
There are still linting issues: https://travis-ci.org/github/release-engineering/pubtools-pulplib/jobs/762457909#L7
It will need some refactoring (and/or change pylint config) that I would rather fix it in different PR.

rohanpm:
> There are still linting issues

I think it's a mistake in the setup of this project, that dependencies are not pinned.  In other projects we've pinned versions of tools such as pylint which means the rules can't suddenly change and break CI.

rohanpm (approved):

PR #94, opened 2021-03-12
Fix pylint and black issues

rbikar:
1) Pinned black and pylint to specific version in tox.ini
2) Add exceptions for pylint related to py3 features
3) Fixed unnecessary creation of list in compat_attr.py

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -96,4 +96,7 @@ disable=print-statement,
         # we use black code style and don't need additional checks
         bad-continuation,
         line-too-long,
+        # skip linting Python 3 features
+        super-with-arguments,
+        raise-missing-from,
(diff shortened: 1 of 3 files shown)


rbikar:
There is still a few pylint issues, but it seems that now it's enough to pass CI.
I tried to fix the duplicated code pulp object models (move repository_memberships attr. to super class), but I hit an issue on py27 test. env.
There was a problem with 'No mandatory attributes allowed after an attribute with a default value or factory.' that I didn't pursue further.
Fore more info see:  https://travis-ci.org/github/release-engineering/pubtools-pulplib/jobs/762620058

rohanpm (approved):

PR #95, opened 2021-06-09
Improved model classes of repository and modulemd

rbikar:
- Added support for several repo notes for Repository model class
  and subclasses.
- Added artifacts and profiles field for ModuleMd unit
- Updated schemas and tests

### DIFF
--- a/pubtools/pulplib/_impl/model/repository/yum.py
+++ b/pubtools/pulplib/_impl/model/repository/yum.py
@@ -91,21 +91,7 @@ class YumRepository(Repository):
         converter=FrozenList,
     )
 
-    arch = pulp_attrib(default=None, type=str, pulp_field="notes.arch")
-    """Repository note describing architecture of the repository"""
-
-    platform_full_version = pulp_attrib(
-        default=None, type=str, pulp_field="notes.platform_full_version"
-    )
-    """Repository note describing full version of the repository"""
-
-    platform_major_version = pulp_attrib(
-        default=None, type=str, pulp_field="notes.platform_major_version"
-    )
-    """Repository note describing major version of the repository"""
-
     ubi_config_version = pulp_attrib(
         default=None, type=str, pulp_field="notes.ubi_config_version"
     )
-    """Repository note describing whatt version (branch) of ubi config should be
-    used for population this repository"""
+    """Version of ubi config that should be used for population of this repository"""
--- a/pubtools/pulplib/_impl/model/repository/yum.py
+++ b/pubtools/pulplib/_impl/model/repository/yum.py
@@ -90,3 +90,22 @@ class YumRepository(Repository):
         type=list,
         converter=FrozenList,
     )
+
+    arch = pulp_attrib(default=None, type=str, pulp_field="notes.arch")
+    """Repository note describing architecture of the repository"""
+
+    platform_full_version = pulp_attrib(
+        default=None, type=str, pulp_field="notes.platform_full_version"
+    )
+    """Repository note describing full version of the repository"""
+
+    platform_major_version = pulp_attrib(
+        default=None, type=str, pulp_field="notes.platform_major_version"
+    )
+    """Repository note describing major version of the repository"""
+
+    ubi_config_version = pulp_attrib(
+        default=None, type=str, pulp_field="notes.ubi_config_version"
+    )
+    """Repository note describing whatt version (branch) of ubi config should be
+    used for population this repository"""
--- a/pubtools/pulplib/_impl/model/unit/modulemd.py
+++ b/pubtools/pulplib/_impl/model/unit/modulemd.py
@@ -81,7 +81,7 @@ class ModulemdUnit(Unit):
     def get_artifacts_filenames(self, exclude_srpms=False):
         """
         Artifacts are typically stored as a list of rpms of nevra format without '.rpm' extension,
-        this method removes the epoch and returns list of actual rpm filenames.
+        this method removes the epoch and returns set of actual rpm filenames.
         """
         regex = r"\d+:"
         reg = re.compile(regex)
--- a/pubtools/pulplib/_impl/model/unit/modulemd.py
+++ b/pubtools/pulplib/_impl/model/unit/modulemd.py
@@ -1,3 +1,5 @@
+import re
+
 from .base import Unit, unit_type
 
 from ..attr import pulp_attrib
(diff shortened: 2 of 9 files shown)


rbikar:
This PR will enable further improvements and refactoring in ubipop.

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -90,3 +90,22 @@ class YumRepository(Repository):
         type=list,
         converter=FrozenList,
     )
+
+    arch = pulp_attrib(default=None, type=str, pulp_field="notes.arch")
+    """Repository note describing architecture of the repository"""
+
+    platform_full_version = pulp_attrib(
```
rohanpm, line 97: I'm not entirely confident about adding these two platform version fields here.

If I remember correctly - these two are quite messy in cdn-utils and it's not really clear what they're supposed to mean.  Docs here don't really help, they just repeat the attribute name without explaining what it means.

Part of the reason to go through this process of modelling all the data into classes is to get away from this legacy with so many poorly defined fields which could contain anything, nobody really knows what they're supposed to be and the only way to figure it out is to read all the code using them.  If we just expose those legacy fields under the same name again without any kind of validation, conversion or explanations for their meaning then we're not improving things.

I really don't want to accept these unless:

- we can provide a proper explanation for what the values are (e.g. not just say it's a full version but explain what exactly that is, x.y.z ?  And explain what is 'platform')
- we can do something to clean up nonsensical values.  e.g. in cdn-utils right now I'm seeing values of '5.9' and $releasever for platform_minor_version.  These seem to be wrong because they aren't minor versions (e.g. in 5.9.0, the minor version would be 9).  I don't want to just "pass on" whatever wrongness already exists.

Can we make it better somehow?  Also if you point at ubipop change which wants to use the fields, I might be able to make some suggestion.

pubtools/pulplib/_impl/model/unit/modulemd.py
```diff
@@ -62,3 +64,32 @@ class ModulemdUnit(Unit):
...
     """
+    artifacts = pulp_attrib(
+        default=None,
+        type=list,
+        converter=frozenlist_or_none_converter,
+        pulp_field="artifacts",
+    )
+    """List of artifacts related to the given module, usually it's a list
+    rpms in nevra format without '.rpm' extension and consist of binary, debug and source
+    rpms as well.
```
rohanpm, line 75: Could we add an example or two in the doc string?

pubtools/pulplib/_impl/model/unit/modulemd.py
```diff
@@ -62,3 +64,32 @@ class ModulemdUnit(Unit):
...
+    """
+
+    profiles = pulp_attrib(type=dict, pulp_field="profiles", default=None)
+    """The profiles of this modulemd unit."""
+
+    def get_artifacts_filenames(self, exclude_srpms=False):
```
rohanpm, line 81: A few suggestions for this one:

- I don't think it should be named with "get".  get prefix shouldn't be needed for a simple accessor I think.  In the existing cases in this library where "get_" is used, the operation involves some remote call to Pulp, which is different to this.
- I don't think the SRPM exclusion should be here. It means the method is doing two unrelated things ("calculate filenames" and "filter according to srpm or not"), and the specific combination is fairly arbitrary.  I guess you added exclude_srpms just because that was your current use-case, but other code might have other filtering needs like picking only certain arches.  Probably it should be left to the caller.
- Once exclude_srpms is removed, `@property` seems to fit well here

So for example, if it's made a `@property` named artifact_filenames then it would be something like this:

```python
# This gives me all the NEVRAs
artifacts = modulemd.artifacts

# If I want filenames instead I do this, which looks similar
rpms = modulemd.artifact_filenames

# And if I don't want SRPMs I can easily filter them myself
rpms = [r for r in rpms if not r.endswith('.src.rpm')]
```

rohanpm (changes requested):

pubtools/pulplib/_impl/model/repository/yum.py
rbikar: Yeah, those fields are not really clear, actually I just want to refactor ubipop a bit and use pulplib Client instead of the custom one in ubipop because the changes for improving performance will be easier when it uses pulplib Client.

In ubipop there are use-cases for those repo notes:
1) Find the counterpart repo of given ubi repo
https://github.com/release-engineering/ubi-population-tool/blob/920b2890aa95a9701288fea2388dd7ea196abefb/ubipop/__init__.py#L325
This can be most probably dropped as we already have 'population_sources' repo note that is calculated in cdn-utils for all ubi repos. 

2) It's used for getting proper branch of ubi-config:
https://github.com/release-engineering/ubi-population-tool/blob/920b2890aa95a9701288fea2388dd7ea196abefb/ubipop/__init__.py#L215 

Basically it needs to know whether the repo is mainline or not and take the proper version of ubi-config either based on ubi_config_version (for mainline repo) or based on platform_full_version for DOT repos. If we could stop populating DOT repos, it would be a way easier. But ideally we should populate 'ubi_config_version' also for DOT repos then we don't need to introduce those repo notes in pulplib.

rohanpm: OK, for (1) I would definitely be in favor of just setting ubi_config_version across *all* repos rather than leaving it blank sometimes and requiring calculation of a default value based on other notes.  If that wouldn't work for some reason, I'd actually prefer to move the calculation of default ubi_config_version into this project rather than have it in ubipop.

For (2) I didn't quite understand it yet, but from your comment it sounds like you're saying that if we populated ubi_config_version everywhere, it would also solve that case.  If that's the case then that sounds like an ideal path forward to me.

pubtools/pulplib/_impl/model/unit/modulemd.py
```diff
@@ -62,3 +64,35 @@ class ModulemdUnit(Unit):
...
+        type=list,
+        converter=frozenlist_or_none_converter,
+        pulp_field="artifacts",
+    )
+    """List of artifacts related to the given module, usually it's a list
+    rpms in nevra format without '.rpm' extension and consist of binary, debug and source
+    rpms as well.
+
+    Example:
+    ["perl-version-7:0.99.24-441.module+el8.3.0+6718+7f269185.src",
```
rohanpm, line 78: Have you checked whether it renders correctly in the docs?

Doesn't really matter though, as I will do some docs revision after this is submitted.

rohanpm (approved):
There is a failing check here from DCO, I assume it was a mistake to enable this tool for the entire organization.  I disabled it and notified the person who installed it.

pubtools/pulplib/_impl/model/unit/modulemd.py
rbikar: I haven't checked that unfortunately :(  
Let's fix it after submission, if it doesn't work.

PR #96, opened 2021-06-10
YumRepository can retrieve related binary, debug or source repo

rbikar:
Added methods to YumRepository that can retrieve
related binary, debug or source repository.
The search is based on relative urls and the expected suffixes
of different yum repositories.

### DIFF
--- a/pubtools/pulplib/_impl/model/repository/yum.py
+++ b/pubtools/pulplib/_impl/model/repository/yum.py
@@ -1,7 +1,13 @@
+import re
+
+from more_executors.futures import f_map, f_proxy
 from .base import Repository, SyncOptions, repo_type
 from ..frozenlist import FrozenList
 from ..attr import pulp_attrib
+from ..common import DetachedException
 from ... import compat_attr as attr
+from ...client.errors import PulpException
+from ...criteria import Criteria
 
 
 @attr.s(kw_only=True, frozen=True)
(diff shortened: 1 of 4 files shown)


pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -90,3 +96,51 @@ class YumRepository(Repository):
         type=list,
         converter=FrozenList,
     )
+
+    @property
+    def related_binary_repository(self):
```
rohanpm, line 101: Overall this is pretty good I think.  Still got a few points:

1. Minor, maybe we can drop the "related" because it's kind of implied.  Of course for every X.some_property, the some_property "relates to" X otherwise you wouldn't be accessing it via X.

2. Could we make it return None (or Future[None]) instead of raising if it doesn't exist?  I know those repos usually occur in (binary, source, debug) triplets but I don't feel confident in saying this should *always* be the case.  It seems to be an unenforced convention.

3.  I think the combination of (property, returns a Future, does not cache the result) should be avoided.  Property gives the impression it's cheap to access the field, but in this case it's not.  We could try to improve that by caching, but honestly I don't know what's the right way to cache the result of a method returning a future.  This is also inconsistent with some of the other properties we have, e.g. rpm_content returns a list and not a future.

Here's a proposal for some guidelines we could stick with to make it consistent:

- If it's a property then it shouldn't return a future.  It's too confusing (e.g. repo.distributors gives you back distributors but repo.related_binary_repository gives you back effectively a *promise to get* a repository, isn't it confusing? Though I know f_proxy helps a little.)
- If it's a property and it involves some remote calls then the outcome should be cached (note: we fail this sometimes now, not saying it has to be fixed here).
- If the purpose of the method is to get some additional data from Pulp and you want it to return a Future, you should name it `get_*`.  This draws the eye to it and highlights it's different from a simple property, for example:

```python
# It's just a property, guess it's a simple value like list or dict
keys = repo.signing_keys

# vs

# hmm, this is named with "get" and is not a property, guess I should pay more
# attention to what it's doing, consider that it might fail, consider caching the result, etc.
keys = repo.get_signing_keys()
```

So if we were to follow above guidelines, I think the choices would be:

A. make it not a property, name it e.g. get_binary_repository, make it return a Future
B. make it a property, name it e.g. binary_repository, make it return non-Future

I suppose that (A) is preferable since it allows using the method in non-blocking code.

rohanpm (changes requested):

pubtools/pulplib/_impl/model/repository/yum.py
rbikar: Yeah, I also wasn't sure about naming and property vs. method problem.
I would go forward with A) suggestion.

rohanpm (changes requested):

pubtools/pulplib/_impl/model/repository/yum.py
```diff
@@ -95,3 +100,47 @@ class YumRepository(Repository):
...
+
+        def unpack_page(page):
+            if len(page.data) != 1:
+                return None
+
+            return page.data[0]
+
+        suffix = suffixes_mapping[repo_t]
+        if str(self.relative_url).endswith(suffix):
+            return self
```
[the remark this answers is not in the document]
rbikar: Done

rohanpm (approved):

PR #99, opened 2021-06-29
Add pulp_repository_published hook using new hooks system

rohanpm:
pubtools recently introduced a hooks/event system. Let's define
a hook invoked after repos are published, for reasons:

- it's a natural way to handle the requirement of "CDN cache should
  be flushed after repo is published"

- as a simple and low-risk first hook to try out the new system.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -100,3 +100,7 @@ disable=print-statement,
         super-with-arguments,
         raise-missing-from,
 
+[TYPECHECK]
+ignored-classes=
+    # dynamic pluggy stuff
+    _HookRelay,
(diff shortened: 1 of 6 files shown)


rbikar (approved):

rohanpm:
I need to hold off on submit of this until python-pubtools RPM is available in eng-rhel-6, else it can interfere with releases.

rohanpm:
> I need to hold off on submit of this until python-pubtools RPM is available in eng-rhel-6, else it can interfere with releases.

About two weeks later, still holding onto it for the same reason.

rohanpm:
Great, that's finally ready, so there should be no release issues with this.

PR #101, opened 2021-07-01
Added GitHub Actions Workflow

daminijain23:
Added .github/workflows/tox-test.yml and .github/workflows/release.yml.
Removed .travis.yml

GitHub Actions and Travis workflows can run parallel. In such case, if both run successfully only then GitHub reports a success.
On reducing the number of tests, coverage also reduced.


.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
```
rohanpm, line 16: Hmm, does this work?  Which Python does it use actually?

If in doubt I'm generally suggesting python 3.8 now, since that's available for RHEL8 and the oldest python3 we'd have to support in most cases.

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,103 @@
+name: Tox tests
+
+on: [push, pull_request]
+
+jobs:
+  py27:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Intsall RPM
```
rohanpm, line 10:
```suggestion
      - name: Install RPM
```

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,103 @@
...
+          python-version: 2.7
+      - name: Install Tox
+        run: pip install tox
+      - name: Run Tox
+        run: tox -e py27
+  static:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Intsall RPM
```
rohanpm, line 24:
```suggestion
      - name: Install RPM
```

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,103 @@
...
+        run: sudo apt-get install -y rpm
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
```
rohanpm, line 29: I suggest bumping all these 3.6 to 3.8, as mentioned above I don't think we need to support any 3.x older than 3.8 now.

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Install Tox
+        run: pip install tox
+      - name: Run Tox
+        run: tox -e static
+  py36-pidiff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Intsall RPM
```
rohanpm, line 38:
```suggestion
      - name: Install RPM
```

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,103 @@
...
+          python-version: 3.6
+      - name: Install Tox
+        run: pip install tox
+      - name: Run Tox
+        run: tox -e pidiff
+  coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Intsall RPM
```
rohanpm, line 52:
```suggestion
      - name: Install RPM
```

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,103 @@
...
+        run: sudo apt-get install -y rpm
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+      - name: Install Tox
+        run: pip install tox
+      - name: Install pytest cov
+        run: pip install pytest-cov
```
rohanpm, line 61: This doesn't seem quite right, because if we already have a "cov" environment in tox.ini (implied by next command "tox -e cov"), then tox should already take care of installing pytest-cov.  So this command should be redundant... can we get rid of it?

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Run Tox
+        run: tox -e cov
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1.5.0
+        with:
+          token: ${{ secrets.CODECOVE_TOKEN }}
```
rohanpm, line 67:
```suggestion
          token: ${{ secrets.CODECOV_TOKEN }}
```

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,103 @@
...
+          fail_ci_if_error: true
+          verbose: true
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Intsall RPM
```
rohanpm, line 74:
```suggestion
      - name: Install RPM
```

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,103 @@
...
+        run: pip install tox
+      - name: Run Tox
+        run: |
+          tox -e docs
+          scripts/push-docs
+  py36-revdep:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Intsall RPM
```
rohanpm, line 90:
```suggestion
      - name: Install RPM
```

rohanpm (changes requested):

.github/workflows/tox-test.yml
daminijain23: Codecov needs coverage library which is installed with pytest-cov. When pytest-cov is installed with tox, it gets installed in a virtual environment and coverage is not available to codecov in next step. 
If we don't install pytest-cov with tox, then it fails to run tests and generate reports, so had to include pytest-cov installation steps in both tox as well as coverage workflow.

.github/workflows/release.yml
daminijain23: Updated to use Python 3.8

.github/workflows/tox-test.yml
daminijain23: Updated workflows for Python 3.8

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,103 @@
...
+        run: sudo apt-get install -y rpm
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install Tox
+        run: pip install tox
+      - name: Run Tox
+        run: tox -e static
+  py36-pidiff:
```
rohanpm, line 34: The name of this (36) doesn't match the python-version used within.

Anyway, I don't think there's a need to run this on more than one python version, so could we just name it 'pidiff' ?

.github/workflows/tox-test.yml
rohanpm: I see, this makes sense, but may I suggest the following then:

- this shouldn't be pytest-cov specifically because codecov-action doesn't depend on "pytest-cov", it only depends on "coverage" (i.e. not the pytest stuff) - right?
- could we please move this to after we run "tox -e cov", i.e. just before we run codecov-action?  I think that makes the intent clearer, that we're installing this just for that action.  A brief comment mentioning that would also be nice.

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,103 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install Tox
+        run: pip install tox
+      - name: Run Tox
+        run: |
+          tox -e docs
+          scripts/push-docs
+  py36-revdep:
```
rohanpm, line 86: 36 in the name here doesn't match the python-version used within.  Anyway, same as for the pidiff case, if the test configuration isn't used on more than one python version then I think we may as well not include any version in the name, so it could just be 'revdep'.

.travis.yml
```diff
@@ -1,39 +0,0 @@
-dist: bionic
```
rohanpm, line 1: To assist with merging this, could you please not delete .travis.yml within this pull request but instead leave it for a later one?

That means we'll be able to do:

- submit this change which activates github-actions
- actions & travis will both temporarily be enabled
- review the behavior of the actions, confirm they're working properly
- do any fixes to actions if needed
- then finally delete travis

If we delete the travis config immediately I'd worry that we'll find out after submitting that the actions don't work, so the repo would have no working CI for a while.

rohanpm (changes requested):

rohanpm:
@daminijain23, I submitted most of the changes here, with commits squashed.  Let's use a new PR if there are further changes needed.

PR #107, opened 2021-08-19
Added vercmp sorting for RpmUnit

rbikar:
Objects of RpmUnit class are now sortable and comparable by standard
vercmp algorithm using epoch, version and release of the unit.
The sorting doesn't take name or arch into account.


rbikar:
Not going forward with this, leaving vercmp sorting in ubipop for now.

PR #113, opened 2021-08-29
ci: bump default py3 version to 3.8

rohanpm:
The primary motivation is to get newer versions of testing tools which
are not available for 3.6. It's also consistent with the github actions
workflows in this repo (which is currently in progress of migrating from
travis to actions).


rohanpm:
I'll make a change to drop travis instead.

PR #117, opened 2021-08-30
ci: retire scripts/push-docs, use action instead

rohanpm:
push-docs does not currently work in github actions because it was coded
to read certain environment variables set by travis.

Replace it with a workflow instead.


rohanpm:
CC @daminijain23

nathanegillett (approved):

PR #121, opened 2021-09-28
Introduce YumRepository.upload_rpm

rohanpm:
Support uploading & importing RPMs into repositories. The API is
consistent with FileRepository.upload_file and some minor refactoring
allows most of the code to be shared between both.

The main source of complexity here is support in the fake client; as
we want to ensure that uploading RPMs via the fake client gives
realistic behavior, we need to parse the RPM headers using behavior
consistent with a real Pulp server.

The upload-files example was also updated to support RPMs.


nathanegillett (approved):

rohanpm:
I don't know why a codecov status didn't appear previously.  I pushed a new commit (with no changes) to see if it'll post a status this time.

rohanpm:
OK, codecov/patch status still won't appear.  Please ignore that, it can be manually checked at https://app.codecov.io/gh/release-engineering/pubtools-pulplib/compare/121/tree if needed.

rbikar (approved):

nathanegillett (approved):

PR #127, opened 2021-10-11
refactor: drop embedded FrozenList

rohanpm:
The FrozenList class here was long ago split out into its own tiny
shared library. Drop the embedded implementation, and the associated
tests, in favor of that library. Related conversion helpers were moved
to the 'convert' module.


pubtools/pulplib/_impl/model/task.py
```diff
@@ -66,11 +67,11 @@ class Task(PulpObject):
     """The ID of the repository associated with this task, otherwise None."""
 
     units = pulp_attrib(
-        default=attr.Factory(FrozenList),
-        type=FrozenList,
+        default=attr.Factory(frozenlist),
+        type=list,
```
MichalHaluza, line 71: This has been refactored from frozenlist to a builtin list. Looks like the change is intentional, but I'm pointing it out just in case it isn't.

rbikar (approved):

rajulkumar (approved):

PR #138, opened 2021-11-25
Add parameters to perform basic authentication with a sync source

querti:
According to docs[1][2], syncing from private container repositories
should be supported by specifying basic auth credentials.

[1] https://docs.pulpproject.org/en/2.11/plugins/pulp_docker/user-guide/release-notes/2.1.x.html
[2] https://docs.pulpproject.org/en/2.11/user-guide/content-sources.html

This change will be necessary for a specific Dynamo rollback edge-case, where we'll want to directly sync certain tags from Quay back to Pulp

Refers to CLOUDDST-10252

### DIFF
--- a/pubtools/pulplib/_impl/model/repository/base.py
+++ b/pubtools/pulplib/_impl/model/repository/base.py
@@ -112,6 +112,14 @@ class SyncOptions(object):
     """A string representing the password that should be used to authenticate with the proxy server
     """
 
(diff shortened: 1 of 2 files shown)


rohanpm:
You've got to bump the version to 2.18.0 in the PR to abide by semver.

rajulkumar (approved):

PR #139, opened 2021-11-26
Added requires and provides attrs to RpmUnit [RHELDST-8603]

rbikar:
Added requires and provides attrs to RpmUnit. Both
are lists of newly introduced class called Dependency.

Requires represent dependencies that RPM requires for installation
and running.

Provides represent capabilities that are provided by the RPM.

### DIFF
--- a/pubtools/pulplib/__init__.py
+++ b/pubtools/pulplib/__init__.py
@@ -12,6 +12,7 @@ from ._impl.model import (
     Unit,
     FileUnit,
     RpmUnit,
+    Dependency,
     ErratumUnit,
     ErratumReference,
     ErratumPackageCollection,
--- a/pubtools/pulplib/__init__.py
+++ b/pubtools/pulplib/__init__.py
@@ -12,7 +12,7 @@ from ._impl.model import (
     Unit,
     FileUnit,
     RpmUnit,
-    Dependency,
+    RpmDependency,
     ErratumUnit,
     ErratumReference,
     ErratumPackageCollection,
--- a/pubtools/pulplib/_impl/fake/rpmlib.py
+++ b/pubtools/pulplib/_impl/fake/rpmlib.py
@@ -25,4 +25,75 @@ except Exception as ex:  # pragma: no cover, pylint: disable=broad-except
     get_header_fields = broken
     get_keys_from_header = broken
 
-__all__ = ["get_rpm_header", "get_header_fields", "get_keys_from_header"]
+__all__ = [
+    "get_rpm_header",
+    "get_header_fields",
+    "get_keys_from_header",
+    "get_rpm_requires",
+    "get_rpm_provides",
+]
+
+
+def get_rpm_requires(header):
+    header_flags_map = {
+        "version": "REQUIREVERSION",
+        "name": "REQUIRENAME",
+        "flags": "REQUIREFLAGS",
+    }
+    return _get_rpm_deps(header, header_flags_map)
+
+
+def get_rpm_provides(header):
+    header_flags_map = {
+        "version": "PROVIDEVERSION",
+        "name": "PROVIDENAME",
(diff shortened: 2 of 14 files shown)


pubtools/pulplib/_impl/model/unit/rpm.py
```diff
@@ -113,6 +137,30 @@ class RpmUnit(Unit):
...
+    requires = pulp_attrib(
+        default=None,
+        type=list,
+        converter=frozenlist_or_none_converter,
+        pulp_field="requires",
+        validator=optional_list_of(Dependency),
+        pulp_py_converter=Dependency._from_data,
+    )
+    """
+    List of capabilities that this RPM provides or ``None`` if this information is unavailable.
```
rohanpm, line 149: The doc strings for requires/provides seem to be the wrong way around.

pubtools/pulplib/__init__.py
```diff
@@ -12,6 +12,7 @@
     Unit,
     FileUnit,
     RpmUnit,
+    Dependency,
```
rohanpm, line 15: Could this class please be named in a way which shows it's RPM-specific?  e.g. RpmDependency?

The concern is that other types of data can also have dependencies and we may need to support those some day, but this Dependency class wouldn't be usable for them.

pubtools/pulplib/_impl/fake/rpmlib.py
```diff
@@ -25,4 +25,75 @@ def broken(*_args, **_kwargs):
...
+    deps = []
+
+    for name, vr, flag in zip(names, versions, flags):
+        version, release = vr.split("-")
+
+        deps_item = {
+            "name": name,
+            "version": version,
+            "release": release,
+            "epoch": None,  # unsure how to get this value from rpm headers
```
rohanpm, line 72: I think this ought to be fixed.  I looked for an RPM with non-zero epoch dependency and I think this info is also in the REQUIREVERSION field?

```
$ rpm -qp FreeWnn-devel-1.10pl020-6.el4.x86_64.rpm --requires
warning: FreeWnn-devel-1.10pl020-6.el4.x86_64.rpm: Header V3 DSA/SHA1 Signature, key ID db42a60e: NOKEY
FreeWnn-libs = 1:1.10pl020-6.el4
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1

$ rpm -qp FreeWnn-devel-1.10pl020-6.el4.x86_64.rpm --queryformat '%{REQUIREVERSION}'
warning: FreeWnn-devel-1.10pl020-6.el4.x86_64.rpm: Header V3 DSA/SHA1 Signature, key ID db42a60e: NOKEY
1:1.10pl020-6.el4
```

So I think the "vr" variable right now is actually "evr", with the "e" component frequently omitted.

rohanpm (changes requested):

rbikar:
Addressed all requested changes + added some more docstrings.

pubtools/pulplib/_impl/model/unit/rpm.py
```diff
@@ -113,6 +157,30 @@ class RpmUnit(Unit):
...
+    requires = pulp_attrib(
+        default=None,
+        type=list,
+        converter=frozenlist_or_none_converter,
+        pulp_field="requires",
+        validator=optional_list_of(RpmDependency),
+        pulp_py_converter=RpmDependency._from_data,
+    )
+    """
+    List of dependecies that this RPM requires or ``None`` if this information is unavailable.
```
rohanpm, line 169:
```suggestion
    List of dependencies that this RPM requires or ``None`` if this information is unavailable.
```

rohanpm (approved):

rbikar:
Fixed typo, I'll squash commits and carry +1s over.

PR #143, opened 2021-12-02
Add some missing pulp_user_metadata fields on files, RPMs

rohanpm:
This adds some fields onto models which will be needed during push but
haven't been included previously.

Note that pulplib will require some updates to support setting these
fields as well, but this is not included here.

### DIFF
--- a/pubtools/pulplib/_impl/model/unit/file.py
+++ b/pubtools/pulplib/_impl/model/unit/file.py
@@ -1,8 +1,11 @@
+import datetime
+
 from .base import Unit, unit_type
 
 from ..attr import pulp_attrib
 from ... import compat_attr as attr
 from ..convert import frozenlist_or_none_sorted_converter
+from ..validate import optional_str, instance_of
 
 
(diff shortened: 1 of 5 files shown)


nathanegillett (approved):

pubtools/pulplib/_impl/model/unit/file.py
```diff
@@ -31,6 +34,46 @@ class FileUnit(Unit):
...
 
+    description = pulp_attrib(
+        type=str,
+        pulp_field="pulp_user_metadata.description",
+        default=None,
+        validator=optional_str,
+    )
+    """A user-oriented terse description of this file.
+
+    .. versionadded:: 2.20.0
```
rohanpm, line 45: I've actually put the wrong version number here when creating the PR, however it might become correct by the time this is submitted, as 2.19.0 seems ready for release.  Maybe ignore this in the review and I'll ensure they're correct at merge time.

rajulkumar (approved):

PR #144, opened 2021-12-02
Introduce Client.search_task to search for Pulp tasks

rajulkumar:
Client has a new search_task api to search for the pulp tasks.
This returns Task objects matching the criteria.

Added the same api to the fake client as well which returns the
tasks added to the controller that match the criteria.
Modified the match_field_eq i.e. EqMatcher for the fake client
to match the values in the list type fields too. It now matches
if one of the values in the list type field is present.

### DIFF
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -355,7 +355,7 @@ class Client(object):
                 Each page will contain a collection of
                 :class:`~pubtools.pulplib.Task` objects.
 
-        .. versionadded:: 2.18.0
+        .. versionadded:: 2.19.0
         """
         return self._search(Task, "tasks", criteria=criteria)
 
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -13,7 +13,7 @@ from six.moves import StringIO
 
 from ..page import Page
 from ..criteria import Criteria
-from ..model import Repository, MaintenanceReport, Distributor, Unit
+from ..model import Repository, MaintenanceReport, Distributor, Unit, Task
 from .search import search_for_criteria
 from .errors import PulpException
 from .poller import TaskPoller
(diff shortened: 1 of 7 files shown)


pubtools/pulplib/_impl/client/client.py
```diff
@@ -340,6 +340,25 @@ def search_distributor(self, criteria=None):
...
+                If None, search for all tasks.
+
+        Returns:
+            Future[:class:`~pubtools.pulplib.Page`]
+                A future representing the first page of results.
+
+                Each page will contain a collection of
+                :class:`~pubtools.pulplib.Task` objects.
+
+        .. versionadded:: 2.18.0
```
rohanpm, line 358: I think this versionadded value is obsolete already, there were probably some releases after you started working on it.

```suggestion
        .. versionadded:: 2.20.0
```

tests/fake/test_fake_search.py
```diff
@@ -430,3 +437,29 @@ def test_search_mapped_field_less_than():
...
+        id="task1",
+        completed=True,
+        succeeded=True,
+        tags=[
+            "pulp:repository:repo1",
+            "pulp:action:publish",
+        ],
+    )
+
+    controller.insert_task(task1)
```
rohanpm, line 455: Could you please make this test also insert some tasks which should NOT be found?

Right now it's a weak test, because e.g. if the fake client search were implemented to ignore the criteria and just return every task every time, the test would still pass.

rohanpm (changes requested):

pubtools/pulplib/_impl/client/client.py
rajulkumar: Thanks. Should the version here be 2.19.0 instead of 2.20.0? I see the latest version as of now is 2.18.0, unless I'm missing anything.

tests/fake/test_fake_search.py
rajulkumar: Agree. Adding a new task here.

pubtools/pulplib/_impl/client/client.py
rohanpm: Yeah, my mistake.  I was thinking the "Added `requires` and `provides`" change had already been released.

rohanpm (approved):

PR #148, opened 2021-12-06
Add unit_id field onto unit models

rohanpm:
Every unit in Pulp has a unique ID. This was not previously exposed as
it seems a good idea to hide that implementation detail if practical.

Add it now to serve the use-case of adjusting pulp_user_metadata fields
on units, for which the API is:

  /pulp/api/v2/content/units/<content_type>/<unit_id>/pulp_user_metadata/

There is no reasonable way to support this API without including unit_id
on the model, so it must be added.

Most of the complexity here is in the fake client, which must generate
unique IDs for units in a deterministic manner so that tests have stable
behavior.

### DIFF
--- a/pubtools/pulplib/_impl/model/unit/erratum.py
+++ b/pubtools/pulplib/_impl/model/unit/erratum.py
@@ -388,3 +388,9 @@ class ErratumUnit(Unit):
     )
     """IDs of repositories containing the unit, or ``None`` if this information is unavailable.
     """
+
+    unit_id = pulp_attrib(type=str, pulp_field="_id", default=None)
+    """The unique ID of this unit, if known.
+
+    .. versionadded:: 2.20.0
+    """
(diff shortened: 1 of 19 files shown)


pubtools/pulplib/_impl/model/unit/erratum.py
```diff
@@ -388,3 +388,9 @@ class ErratumUnit(Unit):
     )
     """IDs of repositories containing the unit, or ``None`` if this information is unavailable.
     """
+
+    unit_id = pulp_attrib(type=str, pulp_field="_id", default=None)
```
rbikar, line 392: I just wonder whether it's possible to have unit_id attr only once in `class Unit` . Or is there some limitation of attrs lib that would disallow proper inheritance?

rohanpm: Yeah, attrs limitation, exactly.  The limitation is that you can't define some attrs without defaults, then some attrs with defaults, then some more attrs without defaults.  It means the base class can't add anything with a default, because the subclasses will then crash when trying to define non-defaults.

Basically if you were to list out all of the attrs from the entire class hierarchy in the order they're defined, all the ones with default values have to appear grouped at the end.

It's because py2 function signatures work this way (can't define some arguments with defaults, then more arguments without defaults afterward).  The limitation goes away when py2 doesn't have to be supported any more.

PR #150, opened 2021-12-08
Add dependencies attr to ModulemdUnit [RHELDST-8604]

rbikar:
Added new attr called dependencies to ModulemdUnit class.
This attr is a list of newly introduced class ModulemdDependency that
represents dependencies (name:stream) of given modulemd unit.

### DIFF
--- a/pubtools/pulplib/_impl/model/unit/modulemd.py
+++ b/pubtools/pulplib/_impl/model/unit/modulemd.py
@@ -1,10 +1,47 @@
 import re
 
-from .base import Unit, unit_type
+from .base import Unit, unit_type, schemaless_init
 
 from ..attr import pulp_attrib
 from ... import compat_attr as attr
 from ..convert import frozenlist_or_none_converter, frozenlist_or_none_sorted_converter
+from ..validate import optional_list_of
+
+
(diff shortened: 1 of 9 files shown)


rajulkumar (approved):

pubtools/pulplib/_impl/model/unit/modulemd.py
```diff
@@ -1,10 +1,47 @@
...
 
 from ..attr import pulp_attrib
 from ... import compat_attr as attr
 from ..convert import frozenlist_or_none_converter, frozenlist_or_none_sorted_converter
+from ..validate import optional_list_of
+
+
+@attr.s(kw_only=True, frozen=True)
+class ModulemdDependency(object):
+    """
```
rohanpm, line 13: Minor: I don't think this shows up in the docs currently, you probably have to add this to docs/api/model.rst.  (I think we missed this for RpmDependency as well)

You can ignore it though, I've been planning for a while to refactor the docs and I might do that soon, as we really shouldn't be dumping so many classes into one big page as we do now anyway.

rohanpm (approved):

PR #156, opened 2022-01-18
Add pulp_repository_pre_publish hookspec [RHELDST-7309]

negillett:
This may be used to subscribe to all repository publish events
triggered by this library.

### DIFF
--- a/pubtools/pulplib/_impl/hooks.py
+++ b/pubtools/pulplib/_impl/hooks.py
@@ -7,13 +7,36 @@ from pubtools.pluggy import hookspec, pm
 
 
 @hookspec
+def pulp_repository_pre_publish(repository, options):
+    """Invoked as the first step in publishing a Pulp repository.
+
+    If a hookimpl returns a non-None value, that value will be used to replace
+    the options for this publish. This can be used to adjust publish options
+    from within a hook.
+
+    Args:
+        repository (:class:`~pubtools.pulplib.Repository`):
+            The repository to be published.
+
+        options (:class:`~pubtools.pulplib.PublishOptions`):
+            The options to use in publishing.
+
+    Returns:
+        options (:class:`~pubtools.pulplib.PublishOptions`):
+            The potentially adjusted options used for this publish.
+    """
+
+
+@hookspec
 def pulp_repository_published(repository, options):
     """Invoked after a Pulp repository has been successfully published.
 
-    :param repository: the repository which has been published.
-    :type repository: pubtools.pulplib.Repository
-    :param options: the options used for this publish.
-    :type options: pubtools.pulplib.PublishOptions
+    Args:
+        repository (:class:`~pubtools.pulplib.Repository`):
+            The repository which has been published.
+
+        options (:class:`~pubtools.pulplib.PublishOptions`):
+            The options used for this publish.
     """
 
 
--- a/pubtools/pulplib/_impl/model/repository/base.py
+++ b/pubtools/pulplib/_impl/model/repository/base.py
@@ -5,8 +5,9 @@ from attr import validators, asdict
 from frozenlist2 import frozenlist
 from more_executors.futures import f_proxy, f_map, f_flat_map
 
-from ..common import PulpObject, Deletable, DetachedException
 from ..attr import pulp_attrib
+from ..common import PulpObject, Deletable, DetachedException
+from ..convert import frozenlist_or_none_converter
 from ..distributor import Distributor
 from ...criteria import Criteria, Matcher
 from ...schema import load_schema
@@ -56,6 +57,15 @@ class PublishOptions(object):
 
     Only relevant if a repository has one or more distributors where
     :meth:`~pubtools.pulplib.Distributor.is_rsync` is ``True``.
+    """
+
+    rsync_extra_args = pulp_attrib(
+        default=None, type=list, converter=frozenlist_or_none_converter
+    )
+    """If present, provide these additional arguments to any rsync commands run
+    during publish.
+
+    Ignored when rsync is not used.
     """
 
 
@@ -382,7 +392,16 @@ class Repository(PulpObject, Deletable):
         if not self._client:
             raise DetachedException()
 
-        # all distributor IDs we're willing to invoke. Anything else is ignored.
+        # Before adding distributors and publishing, we'll activate this hook
+        # to allow subscribing implementers the opportunity to adjust options.
+        hook_rets = pm.hook.pulp_repository_pre_publish(
+            repository=self, options=options
+        )
+        # Use the first non-None hook return value to replace options.
+        hook_rets = [ret for ret in hook_rets if ret is not None]
+        options = hook_rets[0] if hook_rets else options
+
+        # All distributor IDs we're willing to invoke. Anything else is ignored.
         # They'll be invoked in the order listed here.
         candidate_distributor_ids = self._PUBLISH_DISTRIBUTORS
 
@@ -535,11 +554,13 @@ class Repository(PulpObject, Deletable):
     def _config_for_distributor(cls, distributor, options):
         out = {}
 
-        if options.clean is not None and distributor.is_rsync:
-            out["delete"] = options.clean
-
-        if options.origin_only is not None and distributor.is_rsync:
-            out["content_units_only"] = options.origin_only
+        if distributor.is_rsync:
+            if options.clean is not None:
+                out["delete"] = options.clean
+            if options.origin_only is not None:
+                out["content_units_only"] = options.origin_only
+            if options.rsync_extra_args is not None:
+                out["rsync_extra_args"] = options.rsync_extra_args
 
         if options.force is not None:
             out["force_full"] = options.force
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ def get_requirements():
 
 setup(
     name="pubtools-pulplib",
-    version="2.21.0",
+    version="2.22.0",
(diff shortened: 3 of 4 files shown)


pubtools/pulplib/_impl/hooks.py
```diff
@@ -6,14 +6,26 @@
 # pylint: disable=unused-argument
 
 
+@hookspec
+def pulp_repository_pre_publish(to_publish):
```
rohanpm, line 10: I'm not comfortable with some things in the design here:

- the mixing of immutable high-level objects (Distributor) and mutable low-level raw pulp data (override_config) in the same function.
- how different the signature is to the similarly named pulp_repository_published
- the fact that the caller needs to know about the format of override_config, and will then end up relying on pulp2-specific implementation details which we don't document in pulplib.

Could we please make it higher-level instead?  I think the better way might be like this:

- Explicitly add rsync_extra_args to PublishOptions class, with docs along the lines of "If present, provide these additional arguments to any rsync commands run during publish.  Ignored when rsync is not used."
- Implement support internally in pulplib so that if options.rsync_extra_args is set, we adjust override_config to include that for cdn_distributor.
- Make the hook signature like this: `def pulp_repository_pre_publish(repository, options):`
- Define the hook as returning a PublishOptions object.  In the doc string, state that "If a hookimpl returns a non-None value, that value will be used to replace the options for this publish.  This can be used to adjust publish options from within a hook."
- Implement the behavior matching that doc string (i.e. check the return value of the hook)

This keeps us using only the documented high-level API and also gives us a better chance of somehow keeping it working if we move to pulp3.

rohanpm (changes requested):

tests/repository/test_publish.py
```diff
@@ -112,8 +112,14 @@ def test_publish_distributors(fast_poller, requests_mocker, client, hookspy):
...
 
     # It should have invoked hooks
-    assert len(hookspy) == 1
+    assert len(hookspy) == 2
     (hook_name, hook_kwargs) = hookspy[0]
+    assert hook_name == "pulp_repository_pre_publish"
+    assert type(hook_kwargs["to_publish"]) == list
+    assert len(hook_kwargs["to_publish"]) == 2
+    assert type(hook_kwargs["to_publish"][0][0]) == Distributor
+    assert type(hook_kwargs["to_publish"][0][1]) == dict
```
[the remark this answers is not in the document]
negillett: The config part of to_publish is override_config, so here it should be empty dicts.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -58,6 +58,13 @@ class PublishOptions(object):
     :meth:`~pubtools.pulplib.Distributor.is_rsync` is ``True``.
     """
 
+    rsync_extra_args = pulp_attrib(default=None, type=list)
```
rohanpm, line 61: Could you please add converter=frozenlist here to make it immutable?  PublishOptions instances are currently immutable, but this will be lost if we allow a plain mutable list here.

setup.py
```diff
@@ -41,6 +41,11 @@ def get_requirements():
     ],
     install_requires=get_requirements(),
     python_requires=">=2.6",
+    entry_points={
```
rohanpm, line 44: This should not be needed, this entry point is meant to be declared in pubtools-exodus.

pubtools/pulplib/_impl/hooks.py
```diff
@@ -6,14 +6,40 @@
...
 
+@hookspec
+def pulp_repository_pre_publish(repository, options):
+    """Invoked as the first step in publishing a Pulp repository.
+
+    If a hookimpl returns a non-None value, that value will be used to replace
+    the options for this publish. This can be used to adjust publish options
+    from within a hook.
+
+    This should be called before distributors are added, as their config is
```
rohanpm, line 17: I think this sentence about "should be called" is referring to how this hook is meant to be invoked from within pubtools-pulplib, right?

Could we please drop this sentence?  I think the doc strings on the hooks should be targeted only at the hook implementers, not the callers.  For a hook implementer this part of the doc is just confusing, as they are not supposed to be calling this hook ever.  "before distributors are added" is also vague (added to what?)

rohanpm (changes requested):

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -58,6 +58,15 @@ class PublishOptions(object):
     :meth:`~pubtools.pulplib.Distributor.is_rsync` is ``True``.
     """
 
+    rsync_extra_args = pulp_attrib(
+        default=attr.Factory(frozenlist), type=list, converter=frozenlist
```
rohanpm, line 62: The default change I think should be left out, and default=None was better.

The problem with using an empty list here as the default is that it's ambiguous whether empty list means "I don't want to set any rsync_extra_args", or means "I want rsync_extra_args to be set to an empty list".  It should be left as default=None so we can support all these cases:

- None: do not override rsync_extra_args at all, just let the pulp server do what it wants.
- empty list: set rsync_extra_args to an empty list (effectively disabling whatever was the default config in the pulp server).
- non-empty list: set rsync_extra_args to the list I specify (overriding the config in the pulp server).

rohanpm (changes requested):

pubtools/pulplib/_impl/model/repository/base.py
negillett: When default=None, the following error is generated;

```
ImportError while loading conftest '/home/ngillett/src/pubtools-pulplib/tests/conftest.py'.
tests/conftest.py:6: in <module>
    from pubtools.pulplib import Client
pubtools/pulplib/__init__.py:1: in <module>
    from ._impl.client import Client, PulpException, TaskFailedException
pubtools/pulplib/_impl/client/__init__.py:1: in <module>
    from .client import Client
pubtools/pulplib/_impl/client/client.py:14: in <module>
    from ..page import Page
pubtools/pulplib/_impl/page.py:10: in <module>
    from .model.attr import pulp_attrib
pubtools/pulplib/_impl/model/__init__.py:2: in <module>
    from .repository import (
pubtools/pulplib/_impl/model/repository/__init__.py:1: in <module>
    from .base import Repository, PublishOptions, SyncOptions
pubtools/pulplib/_impl/model/repository/base.py:132: in <module>
    class Repository(PulpObject, Deletable):
pubtools/pulplib/_impl/model/repository/base.py:366: in Repository
    def publish(self, options=PublishOptions()):
E   TypeError: 'NoneType' object is not iterable
ERROR: InvocationError for command /home/ngillett/src/pubtools-pulplib/.tox/py36/bin/pytest -v (exited with code 4)
```
This seems to be due to the validator attached by pulp_attrib() because...
```
>>> isinstance(None, list)
False
```
I noticed that there's no attr defined in repository.base with default=None so, thinking this error may be a factor, I followed suit.
Maybe this necessitates a change to pulp_attrib(), though...

negillett:
This is now blocking https://github.com/release-engineering/pubtools-exodus/pull/2, as it depends on the new PublishOption.rsync_extra_args.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -58,6 +58,13 @@ class PublishOptions(object):
     :meth:`~pubtools.pulplib.Distributor.is_rsync` is ``True``.
     """
 
+    rsync_extra_args = pulp_attrib(default=None, type=frozenlist)
```
negillett, line 61: Setting `type=frozenlist` seems to be the only way to make the list immutable and permit `default=None`.

rohanpm: This won't make the list immutable because type doesn't really do anything, it's basically just for documentation purposes.  It doesn't actually do a conversion.

The problem is that my suggestion of converter=frozenlist was not correct if we want to allow None values.  Looking at other similar cases in the classes in this library, we use frozenlist_or_none_converter.

For example ModulemdUnit.artifacts is defined like this:

```python
    artifacts = pulp_attrib(
        default=None,
        type=list,
        converter=frozenlist_or_none_converter,
        pulp_field="artifacts",
    )
```

So we want the same thing here except we don't need `pulp_field`.

rohanpm (changes requested):

tests/repository/test_publish.py
```diff
@@ -146,7 +152,9 @@ def test_publish_with_options(requests_mocker, client):
         ],
     )
 
-    options = PublishOptions(clean=True, force=True, origin_only=True)
+    options = PublishOptions(
+        clean=True, force=True, origin_only=True, rsync_extra_args=frozenlist(["-a"])
```
rohanpm, line 156: Can you please not explicitly initialize it with a frozenlist?

We want it to work with just rsync_extra_args=["-a"] and for the conversion to a frozenlist to be a hidden implementation detail.  Initializing without `frozenlist` would be the typical case, so that's what the test should do.

rohanpm (changes requested):

tests/repository/test_publish.py
negillett: Ah, I missed this. Yeah, it should just be a list.

tests/repository/test_publish.py
```diff
@@ -1,6 +1,8 @@
 import logging
 import pytest
 
+from frozenlist2 import frozenlist
```
rohanpm, line 4: Minor - this is unused now.

rohanpm (approved):

rohanpm (approved):

PR #159, opened 2022-02-04
Refresh black version & setup

rohanpm:
This change consists of the following:

- drop black from 'tox -e static' because pre-commit.ci has been enabled
  for this repo. It's redundant to have it run from two places, and
  broken if they don't use exactly the same version.

- update black version in pre-commit config via pre-commit autoupdate

- reformat some code according to the desired behavior from the new
  version of black


rohanpm:
There are problems with unicode string literals since we still need to support python 2.  Closing this for now, may revisit later.

PR #162, opened 2022-02-21
Enable searching for provides and requires attrs in fake client

rbikar:
Real pulp server can be queried for units by subfield like 'provides.name'
that matches subfield in a list of objects stored in given attr.

Let's add this capability to the fake client.

The change affects only queries related to provides or requires attrs on
RpmUnit.

### DIFF
--- a/pubtools/pulplib/_impl/fake/match.py
+++ b/pubtools/pulplib/_impl/fake/match.py
@@ -144,6 +144,13 @@ def pulp_value(pulp_field, obj):
     for cls_field in attr.fields(type(obj)):
         pulp_key = cls_field.metadata.get(PULP2_FIELD)
         if pulp_key:
+            subfield = None
+            # provides and requires attributes of RpmUnit needs special conversion
+            # method as we want to get values of specific attribute in list of objects
+            if pulp_key == "provides" or pulp_key == "requires":
+                pulp_key = pulp_field
+                subfield = pulp_field.split(".")[1]
+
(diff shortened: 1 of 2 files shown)


rajulkumar (approved):

pubtools/pulplib/_impl/fake/match.py
```diff
@@ -144,25 +144,33 @@ def pulp_value(pulp_field, obj):
     for cls_field in attr.fields(type(obj)):
         pulp_key = cls_field.metadata.get(PULP2_FIELD)
         if pulp_key:
+            subfield = None
+            # provides and requires attributes of RpmUnit needs special conversion
+            # method as we want to get values of specific attribute in list of objects
+            if pulp_key == "provides" or pulp_key == "requires":
```
rohanpm, line 150: I'm just curious if you know of anything which would make this not work for other fields as well?  e.g. if the condition was just `if "." in pulp_field` so that it would try to process any nested field and not treat these two as a special case.

However, restricting the behavior only to those fields where we know it works (and need it to work) is certainly valid.

rohanpm (approved):

pubtools/pulplib/_impl/fake/match.py
rbikar: Actually I'm not sure about that. But I guess I'll stick with this solution for the time being. We can improve this in future, if we need to use fake client for queries for other nested fields, similar to provides and requires.

PR #163, opened 2022-02-21
Deprecate support for legacy sigstore [CLOUDDST-5513]

neithere:
This change deprecates the repository attribute `is_sigstore` as the signatures are not stored in a Pulp repository any more.

The next minor version 2.24 appears suitable for deprecation.

### DIFF
--- a/pubtools/pulplib/_impl/model/repository/file.py
+++ b/pubtools/pulplib/_impl/model/repository/file.py
@@ -40,6 +40,12 @@ class FileRepository(Repository):
         type=bool,
         validator=validators.instance_of(bool),
     )
+    """True if this is a sigstore repository, used for container image manifest
+    signatures.
+
+    .. deprecated:: 2.24.0
+       The signatures are not stored in a Pulp repository any more.
+    """
 
     mutable_urls = attr.ib(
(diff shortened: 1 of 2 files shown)


pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -40,6 +40,12 @@ class FileRepository(Repository):
         type=bool,
         validator=validators.instance_of(bool),
     )
+    """True if this is a sigstore repository, used for container image manifest
```
rohanpm, line 43: Can you please leave this one out?  The convention here is not to duplicate doc strings in the case where an attribute is being inherited from the base class.  This is why for example "type" and "mutable_urls" have no doc string.

rohanpm (changes requested):

pubtools/pulplib/_impl/model/repository/file.py
neithere: Sure, will update. Thanks!

rohanpm (approved):

rajulkumar (approved):

midnightercz (approved):

rbikar (approved):

PR #166, opened 2022-02-23
Added filters to remove content from a repository

rajulkumar:
Repository.remove_content filters content only on type_ids.
This patch adds the option to provide `filters` which will
filter the contents for removal based on the criteria provided
for the given `type_ids`.


pubtools/pulplib/_impl/fake/client.py
```diff
@@ -491,17 +491,16 @@ def _do_unassociate(self, repo_id, criteria=None):
         removed_units = set()
         kept_keys = set()
 
+        criteria = criteria or Criteria.true()
+        # only validating the criteria here like in actual scenario.
         # use type hint=Unit so that if type_ids are the goal here
```
rohanpm, line 496: I think this old comment is obsolete now, isn't it?

It was saying that we call search_for_criteria just to get a pulp_search for which we can look at type_ids.  But your new comment says we are just doing this to validate the criteria object, which I think is valid, meaning the old comment can be removed.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -479,6 +479,11 @@ def remove_content(self, **kwargs):
 
                 If omitted, content of all types will be removed.
 
+            filters (:class:`~pubtools.pulplib.Criteria`)
```
rohanpm, line 482: This should be called `criteria` and not `filters` - that's what it's called for all the similar methods like search_content, copy_content.

Also could you please document this as the first argument?  I explained a bit more in another comment, but summarized - the idea was that, once this function supports a Criteria object, then that's the preferred way of using it.  type_ids might eventually be deprecated and removed since we don't really need both.  As the most important/preferred way of using the method now, it'd be best to mention the criteria arg first.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -525,6 +531,11 @@ def remove_content(self, **kwargs):
                 "_content_type_id",
                 Matcher.in_(type_ids),  # Criteria.with_field_in is deprecated
             )
+        # extend the criteria with the filters when available.
+        # search_for_criteria will fill out the PulpSearch#filters field with
+        # the rest of the content excluding type_ids
+        if filters is not None:
+            criteria = Criteria.and_(criteria, filters)
```
rohanpm, line 538: Does this work if type_ids was not provided?  If so, criteria is going to be `None` at this point, so this will be equivalent to `Criteria.and_(None, filters)`.  Not sure and_ is designed to accept `None`.

I think that case should be tested - actually I think that should be the preferred usage, because Criteria already supports matching by type (Criteria.with_unit_type).  So with this change, the caller can use just the criteria object to achieve everything and there's no need to pass type_ids separately.

I believe the only reason type_ids was there is because it was required for the first removal use-case, but full Criteria support was not required yet when remove_content was introduced.  So it was added to use type_ids, but via *kwargs to ensure it could be deprecated at a later point when full criteria support was added as you're doing now.

tests/fake/test_fake_remove_content.py
```diff
@@ -100,3 +106,77 @@ def test_remove_deleted_repo_fails():
...
+        ),
+    ]
+    units = rpm_units + modulemd_units
+
+    repo = YumRepository(id="repo1")
+    controller.insert_repository(repo)
+    controller.insert_units(repo, units)
+
+    filters = Criteria.with_field("name", "bash")
+    remove_rpms = client.get_repository("repo1").remove_content(
```
rohanpm, line 135: Could we please have at least one case which calls remove_content with criteria and without type_ids ?  (which I'm saying elsewhere should now be the preferred usage)

tests/repository/test_remove_content.py
```diff
@@ -171,3 +173,47 @@ def test_remove_loads_units(fast_poller, requests_mocker, client):
...
+    requests_mocker.post(
+        "https://pulp.example.com/pulp/api/v2/tasks/search/",
+        [{"json": [{"task_id": "task1", "state": "finished"}]}],
+    )
+    filters = Criteria.and_(
+        Criteria.with_field("filename", "some.rpm"),
+        Criteria.with_field("signing_key", Matcher.in_(["key123"])),
+    )
+
+    assert repo.remove_content(
```
rohanpm, line 197: Could we please have at least one case which calls remove_content with criteria and without type_ids ?

rohanpm (changes requested):

tests/repository/test_remove_content.py
rajulkumar: As per the documents and the bug https://bugzilla.redhat.com/show_bug.cgi?id=1021579, type_ids should be present for the filters in criteria to be considered. If type_ids is omitted, filters are skipped and all the content is removed. I have updated the test with this assumption that type_ids will be there with the filters in the criteria.
Should there be an error thrown from _do_unassociate if there's a filter but no type_id OR the mention in the doc is fine?

pubtools/pulplib/_impl/model/repository/base.py
rajulkumar: I have removed the filters and updated it with criteria. Criteria will be the one taken into account and will ignore type_ids kwarg if criteria is present.

pubtools/pulplib/_impl/fake/client.py
rajulkumar: Agree. Updated the comment.

pubtools/pulplib/_impl/model/repository/base.py
rajulkumar: Done.

tests/repository/test_remove_content.py
rohanpm: It seems a bad idea to go ahead and remove everything if there were no type_ids.  I see two options:

1. In do_unassociate, return an error if the search had no type_ids.  The error should be worded in terms of the remove_content API, e.g. "Criteria to remove_content must specify at least one unit type!"
2. If there are no type_ids, then just before submitting the search, set them to self.get_content_type_ids() (i.e. default to all the type ids)

(1) feels like the safer approach to me, as it's dangerous anyway to do a removal without specifying the type(s).

rajulkumar: Updated as per option(1) for both the client and fake_client

rohanpm (approved):

PR #168, opened 2022-03-02
tests: drop the verylarge upload test due to mem usage issues

rohanpm:
This test was added recently due to some earlier performance issues with
upload. Unfortunately, this test itself can trigger some performance
issues because of the way requests_mocker works.

The test uploads 2GB worth of 0 bytes. The problem is that
requests-mock keeps a reference to every request body as part
of that upload. It does this for (at least) the sake of request_history.
Therefore the test really needs >2GB of memory even though the code
under test does not. I wasn't aware of this behavior from requests-mock
when the test was written.

Unfortunately there's no way to opt-out of this behavior in
requests-mock (at least not using public API), so there is no simple
way how to fix the test. Let's drop it as it does not seem appropriate
to allow for such high memory usage from one test.

CI for this repo has started to become a bit flaky recently (some
timeouts). I'm not certain, but it's possible this was a cause.


rohanpm:
I meant to create this as Draft but misclicked.  I'm going to close it as I can see already that it hasn't fixed the CI problem.

PR #169, opened 2022-03-02
Add test timeouts & verbose args for debugging

rohanpm:
Trying to figure out the cause of hangs in CI.


rohanpm:
Guess I'll abandon this as those CI issues seem to have gone away mysteriously...

PR #182, opened 2022-04-01
Fix paginated searches sometimes being cancelled incorrectly

rohanpm:
All Pulp searches are paginated. Large searches can potentially take
several minutes to iterate through all pages, so it's worthwhile trying
to stop those searches if we can somehow determine that the caller is no
longer interested in the result.

This was previously done by using a weakref which will trigger
cancellation of the next page if the current page is garbage collected.
However, the logic here was flawed: just because the self is
being GC'd, does not guarantee that someone hasn't already accessed
self.next and attached some callback to it. This meant it was
possible for searches to be wrongly cancelled even if someone was still
interested in the result.

Revise it to fix this. If 'next' is accessed at all after construction,
we assume it's no longer safe to cancel.

There probably isn't a perfect way to do this with the current API, but
this commit should at least swing the pendulum back from unsafe
behavior "it sometimes cancels when it shouldn't", to the safe but
less efficient behavior "it sometimes doesn't cancel when it
theoretically could".

### DIFF
--- a/pubtools/pulplib/_impl/page.py
+++ b/pubtools/pulplib/_impl/page.py
@@ -87,22 +87,31 @@ class Page(object):
 
     def __attrs_post_init__(self):
         if self.next:
-            # If *this* page disappears, it's pointless to keep querying for
-            # the next page, so we could cancel that future.
+            # We have a next, which means there are subsequent pages currently being
+            # searched for.
+            #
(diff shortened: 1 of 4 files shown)


rohanpm:
Note, the CI failure would be fixed by https://github.com/release-engineering/pubtools-pulplib/pull/183.

rbikar (approved):

PR #190, opened 2022-07-01
Expose include_in_download_service repo notes

MichalHaluza:
Recently emerging utilities developed and used by release engineers
try to favor modern libraries such as pubtools-pulplib over the legacy
alternatives, however some functions existing in older libraries are
not provided by this library yet. The possibility to set or read
include_in_download_service[_preview] repo notes is one such feature
currently required, so this commit makes the repo notes accessible.

### DIFF
--- a/pubtools/pulplib/_impl/model/repository/base.py
+++ b/pubtools/pulplib/_impl/model/repository/base.py
@@ -303,6 +303,34 @@ class Repository(PulpObject, Deletable):
     .. versionadded:: 2.29.0
     """
 
+    include_in_download_service = pulp_attrib(
+        default=False,
+        type=bool,
+        mutable=True,
+        pulp_field="notes.include_in_download_service",
+        pulp_py_converter=lambda x: x == "True",
+        py_pulp_converter=str,
+    )
+    """Flag indicating whether the repository is visible in production instance
+    of download service.
+
+    .. versionadded:: 2.34.0
+    """
+
+    include_in_download_service_preview = pulp_attrib(
+        default=False,
+        type=bool,
+        mutable=True,
+        pulp_field="notes.include_in_download_service_preview",
+        pulp_py_converter=lambda x: x == "True",
+        py_pulp_converter=str,
+    )
+    """Flag indicating whether the repository is visible in staging instance
+    of download service.
+
+    .. versionadded:: 2.34.0
+    """
+
     @distributors.validator
     def _check_repo_id(self, _, value):
         # checks if distributor's repository id is same as the repository it
--- a/pubtools/pulplib/_impl/schema/repository.yaml
+++ b/pubtools/pulplib/_impl/schema/repository.yaml
@@ -102,6 +102,20 @@ properties:
       ubi_config_version:
         type: string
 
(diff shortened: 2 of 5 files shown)


pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -303,6 +303,36 @@ class Repository(PulpObject, Deletable):
     .. versionadded:: 2.29.0
     """
 
+    include_in_download_service = pulp_attrib(
```
rohanpm, line 306: The commit message mentioned the need to set the field, but if you want to be able to do it with this library you'll have to set mutable=True in here.  See `product_versions` field as an example.  Do you want to include that in this change?

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -303,6 +303,36 @@ class Repository(PulpObject, Deletable):
...
+        default=False,
+        type=bool,
+        pulp_field="notes.include_in_download_service",
+        pulp_py_converter=lambda x: x == "True",
+        py_pulp_converter=str,
+    )
+    """Flag indicating whether the repository is visible in production instance
+    of download service.
+
+    Although the value looks like a boolean, it is always stored as string in Pulp.
```
rohanpm, line 316: I'd rather not even mention this as the library is supposed to encapsulate entirely how the fields are serialized in Pulp.  If you stick with this library you'll never even see this quirk.  In the doc strings for other fields we don't generally document how they are being serialized (e.g. signing_keys, product_versions).

rohanpm (commented):
For the CI failures:

- for docs, it will be fixed by https://github.com/release-engineering/pubtools-pulplib/pull/191
- for the pidiff failure, you've got to bump the version in setup.py so the library follows SemVer rules.

pubtools/pulplib/_impl/model/repository/base.py
MichalHaluza: The comment should have served as an explanation for the next person reading the codebase and thinking "_Why is a boolean value interpreted as a string? This is surely a mistake. Let me fix it._". On the other hand you're right that the lists you mention are also stored as json strings without anyone raising an eyebrow (although in the boolean case the eyebrow should be risen much higher, especially because we have also other boolean values with are proper booleans)

MichalHaluza:
I've expected the version bump to happen during release, same as changelog update. Should I update also the changelog manually as part of this PR?

rohanpm:
> I've expected the version bump to happen during release, same as changelog update.

The reason it's enforced earlier is that, if happening at release, it requires the person doing the release to evaluate what's in the release and how the SemVer rules apply.  I don't trust that this would happen correctly.  Also if someone accidentally makes a backwards-incompatible change, it should be found immediately and not later at release time.  Version bump does still have to happen at release time if it hasn't already happened earlier though.

> Should I update also the changelog manually as part of this PR?

It's optional... the original idea was that yes, you would, but it's mostly been dropped until there is a solution to "how would you prevent everyone's PR conflicting with each other then".

rohanpm (approved):

PR #193, opened 2022-07-13
Fixed converting cdn_published criteria search in RpmUnit and FileUnit 

jajreidy:
This will fix a bug when running a criteria search against cdn_published. Cdn_published is stored as a string but criteria is trying to use $date in the search.

### DIFF
--- a/pubtools/pulplib/_impl/model/convert.py
+++ b/pubtools/pulplib/_impl/model/convert.py
@@ -66,6 +66,17 @@ def tolerant_timestamp(value):
     return value
 
 
+def timestamp_converter(value):
+    # Converter for fields which are stored as strings,
+    # but which model is expecting datetime
+    # falls back to returning the input verbatim if not a datetime.
+    #
+    if isinstance(value, datetime.datetime):
+        return value.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+    return value
+
+
 def write_timestamp(value):
     # defaults to current time if value is None
     if value is None:
--- a/tests/client/test_search.py
+++ b/tests/client/test_search.py
@@ -8,6 +8,8 @@ from pubtools.pulplib._impl.client.search import (
     field_match,
 )
 
+from pubtools.pulplib import RpmUnit, FileUnit
+
 
 def test_null_criteria():
     """Searching for None or True translates to empty filters."""
@@ -99,4 +101,36 @@ def test_dict_matcher_value():
 
     assert filters_for_criteria(crit) == {
         "created": {"$lt": {"created_date": {"$date": "2019-09-04T00:00:00Z"}}}
+    }
+
+
+def test_cdn_published():
+    """criteria using cdn_published as matcher value
+    with a datetime in RpmUnit, FileUnit
+    """
+
+    crit = Criteria.with_field("cdn_published", datetime.datetime(2019, 9, 4, 0, 0, 0))
+
+    assert filters_for_criteria(crit, RpmUnit) == {
+        "pulp_user_metadata.cdn_published": {"$eq": "2019-09-04T00:00:00Z"}
+    }
+
+    assert filters_for_criteria(crit, FileUnit) == {
+        "pulp_user_metadata.cdn_published": {"$eq": "2019-09-04T00:00:00Z"}
+    }
+
+
+def test_cdn_published_not_datetime():
+    """criteria using cdn_published as matcher value
+    with a string in RpmUnit, FileUnit
+    """
+
+    crit = Criteria.with_field("cdn_published", "not datetime")
+
+    assert filters_for_criteria(crit, RpmUnit) == {
+        "pulp_user_metadata.cdn_published": {"$eq": "not datetime"}
+    }
+
+    assert filters_for_criteria(crit, FileUnit) == {
+        "pulp_user_metadata.cdn_published": {"$eq": "not datetime"}
     }
--- a/pubtools/pulplib/_impl/model/unit/file.py
+++ b/pubtools/pulplib/_impl/model/unit/file.py
@@ -4,7 +4,11 @@ from .base import Unit, unit_type
 
 from ..attr import pulp_attrib
 from ... import compat_attr as attr
-from ..convert import frozenlist_or_none_sorted_converter, tolerant_timestamp
+from ..convert import (
+    frozenlist_or_none_sorted_converter,
+    tolerant_timestamp,
+    timestamp_converter,
+)
 from ..validate import optional_str, instance_of
 
 
@@ -125,6 +129,7 @@ class FileUnit(Unit):
         pulp_field="pulp_user_metadata.cdn_published",
         default=None,
         converter=tolerant_timestamp,
+        py_pulp_converter=timestamp_converter,
         validator=instance_of((datetime.datetime, type(None))),
     )
     """Approximate :class:`~datetime.datetime` in UTC at which this file first
--- a/pubtools/pulplib/_impl/model/unit/rpm.py
+++ b/pubtools/pulplib/_impl/model/unit/rpm.py
@@ -9,6 +9,7 @@ from ..convert import (
     frozenlist_or_none_converter,
     frozenlist_or_none_sorted_converter,
     tolerant_timestamp,
+    timestamp_converter,
 )
 from ..validate import optional_str, instance_of
 from ..common import PulpObject
@@ -181,6 +182,7 @@ class RpmUnit(Unit):
         pulp_field="pulp_user_metadata.cdn_published",
         default=None,
         converter=tolerant_timestamp,
+        py_pulp_converter=timestamp_converter,
         validator=instance_of((datetime.datetime, type(None))),
     )
     """Approximate :class:`~datetime.datetime` in UTC at which this RPM first


rbikar:
> This will bring Matcher.less_than more in line with how less_than works with RpmUnits in pub.

here, we are talking about units in pulp, not pub. Please fix the commit message and PR descr. as it's now misleading.

pubtools/pulplib/_impl/fake/match.py
```diff
@@ -131,10 +132,26 @@ def match_in(matcher, field, obj):
 
 @visit(LessThanMatcher)
 def match_field_less(matcher, field, obj):
+    if isinstance(matcher._value, six.string_types):
```
rbikar, line 135: This now make behavior inconsistent if someone uses matcher that goes via match_field_eq(), can you apply the same there?

rbikar (dismissed):

jajreidy:
> > This will bring Matcher.less_than more in line with how less_than works with RpmUnits in pub.
> 
> here, we are talking about units in pulp, not pub. Please fix the commit message and PR descr. as it's now misleading.

Done and done

pubtools/pulplib/_impl/fake/match.py
```diff
@@ -178,7 +203,7 @@ def convert_field_to_pulp(obj, field_name, value, subfield):
     if cls_field.type is datetime.datetime and isinstance(value, datetime.datetime):
         # We can convert from datetime back to the ISO8601 timestamp format
         # used within Pulp
-        return value.strftime("%Y-%m-%dT%H:%M:%SZ")
+        return value
```
rajulkumar, line 206: Either we should get rid of the if condition here, as 'value' is anyways being returned in the end OR avoid converting the matcher.\_value in the match\_* methods above.
I'll prefer the later and keeping this as-is to keep the matcher.\_value consistent with the definition rather than modifying it somewhere midway.
I suggest the transformed datetime object in the match\_* above is used there as required. That's the only place the transformed value is required as of now. Updating it to matcher.\_value may only add confusion as same object when passed through different matcher has a different value e.g. match\_field\_less changes the value but match\_in doesn't.
Also, the pulp\_dict in pulp\_value() looks more in line with how the objects exist in Pulp and this will keep it consistent.

rajulkumar (changes requested):

pubtools/pulplib/_impl/fake/match.py
jajreidy: The problem ran into was convert_field_to_pulp was converting a datetime to a string when if you access the field directly it returns it as datetime. Creating some weirdness between "notes.created" and "created_date", "notes.created" was returning a string where "created_date" was returning a datetime.

We tested this against the real Matcher in pulp to ensure that this actually represents what is going on in pulp. I agree removing the if statement entirely though (because it doesn't make sense to have it there).

rajulkumar: I see what you mean. That's how Pulp works as well, converting the str to datetime and compare if provided the $date operator, which we usually do with datetime type values.
So, It's the first one then, getting rid of the condition :)
I'll suggest to add this transformation to match\_in matcher as well just to keep it consistent, though we don't have a use case yet matching multiple datetime values.

jajreidy: All fixed up :)
Yeah I don't want this code to be this way.. but it matches what was seen in testing :(

pubtools/pulplib/_impl/fake/match.py
```diff
@@ -96,6 +97,12 @@ def match_field(criteria, obj):
 
 @visit(EqMatcher)
 def match_field_eq(matcher, field, obj):
+    if isinstance(matcher._value, six.string_types):
```
rohanpm, line 100: A couple of issues with this:

1. I don't think applying a matcher should have the side-effect of mutating it (i.e. this function should not be modifying matcher._value).  If this logic is otherwise correct, I think it should be implemented in a way which doesn't modify any of the passed in objects, like

```
# By default we are using exactly the matcher's stored value
to_match = matcher._value

# But under certain circumstances we might use a different value
if some_conditions(...):
  to_match = something_else(...)

# Now use the value however it's appropriate to do so
if value_matches(to_match):
    ...
```

2. A bigger problem is that the logic just doesn't seem to make sense: it wouldn't be right to assume that, *whenever* a string can be parsed as a datetime, then it's correct to do so.  Strings which look like datetimes could also be stored in actual string fields in Pulp, and then the logic here is going to mess up some matches on those fields.  cdn_published should be an example of this.

rohanpm (changes requested):
Could you please explain what problem is meant to be solved by this?

Given the PR is linked from RHELDST-4004 I'd have expected it's supposed to fix the testing failure found in that issue.  But that doesn't seem to make sense as you're only changing the behavior of the fake client here, while the tester reported an issue testing against a real Pulp server where none of this code should have an impact.

pubtools/pulplib/_impl/fake/match.py
rohanpm: Actually, cdn_published wasn't a good example as it'll really be stored as a datetime when using fake client.  erratum updated/issued fields are a better example, as they are defined as string timestamps of no strictly enforced format.  So they can happen to contain values matching the logic here and be parsed into timestamps when it's not appropriate to do so.

jajreidy: I was trying to match what I saw what the real less_than matcher does compared to the fake matcher. I agree the logic doesn't make sense but I pulled that line of code from the pubtools-pulp code :\

What I've seen is that if you pass datetime it converts the $lt search to YYYY-MM-DDTHH:MM:SSZ but items in search are set to YYYY-MM-DD HH:MM:SS. I can't just change the way the datetime is converted or send a string since the tests expect a datetime. I don't really feel comfortable with changing any core functionality in pulp to fix the issue.

here in target_pulp2.py this is what it sets cdn_published to cdn_published = datetime.datetime.now().strftime('%Y-%m-%dT%H:%M:%S%z')

rohanpm: Sorry, I'm still not getting it.

> if you pass datetime it converts the $lt search to YYYY-MM-DDTHH:MM:SSZ

Right, this is $date operator implemented by Pulp, mentioned at https://docs.pulpproject.org/en/2.19/dev-guide/conventions/criteria.html#search-criteria .

> but items in search are set to YYYY-MM-DD HH:MM:SS

What are "items in search" and what do you mean by "set to"?

Could you please reply to https://github.com/release-engineering/pubtools-pulplib/pull/193#pullrequestreview-1048728884 ?

jajreidy:
> Could you please explain what problem is meant to be solved by this?
> 
> Given the PR is linked from RHELDST-4004 I'd have expected it's supposed to fix the testing failure found in that issue. But that doesn't seem to make sense as you're only changing the behavior of the fake client here, while the tester reported an issue testing against a real Pulp server where none of this code should have an impact.

With testing with Robert we found that running this search:


        criteria = Criteria.and_(
            Criteria.with_unit_type(RpmUnit),
            Criteria.with_field(
                "cdn_published",
                Matcher.less_than((datetime.utcnow() - timedelta(days=arc_threshold)).strftime('%Y-%m-%dT%H:%M:%S%Z')),
            ),
        )
       returned multiple responses versus:
       
        criteria = Criteria.and_(
            Criteria.with_unit_type(RpmUnit),
            Criteria.with_field(
                "cdn_published",
                Matcher.less_than((datetime.utcnow() - timedelta(days=arc_threshold))),
            ),
        )
        
        wouldn't get any responses. So this was just making the fake client a little less strict but I'm not even 30% sure it's the right direction.

rohanpm:
> With testing with Robert we found that running this search:
> 
> ```
>     criteria = Criteria.and_(
>         Criteria.with_unit_type(RpmUnit),
>         Criteria.with_field(
>             "cdn_published",
>             Matcher.less_than((datetime.utcnow() - timedelta(days=arc_threshold)).strftime('%Y-%m-%dT%H:%M:%S%Z')),
>         ),
>     )
>    ```
>    returned multiple responses versus:
>    ```
>     criteria = Criteria.and_(
>         Criteria.with_unit_type(RpmUnit),
>         Criteria.with_field(
>             "cdn_published",
>             Matcher.less_than((datetime.utcnow() - timedelta(days=arc_threshold))),
>         ),
>     )
>    ```
>  wouldn't get any responses.

Got it.  I think the correct fix is along these lines (just with testing added & probably using a function in 'convert' module rather than a duplicated lambda):

```
--- a/pubtools/pulplib/_impl/model/unit/file.py
+++ b/pubtools/pulplib/_impl/model/unit/file.py
@@ -125,6 +125,9 @@ class FileUnit(Unit):
         pulp_field="pulp_user_metadata.cdn_published",
         default=None,
         converter=tolerant_timestamp,
+        py_pulp_converter=lambda value: value.strftime("%Y-%m-%dT%H:%M:%SZ")
+        if isinstance(value, datetime.datetime)
+        else value,
         validator=instance_of((datetime.datetime, type(None))),
     )
     """Approximate :class:`~datetime.datetime` in UTC at which this file first
diff --git a/pubtools/pulplib/_impl/model/unit/rpm.py b/pubtools/pulplib/_impl/model/unit/rpm.py
index c4ec9ce..61d063e 100644
--- a/pubtools/pulplib/_impl/model/unit/rpm.py
+++ b/pubtools/pulplib/_impl/model/unit/rpm.py
@@ -181,6 +181,9 @@ class RpmUnit(Unit):
         pulp_field="pulp_user_metadata.cdn_published",
         default=None,
         converter=tolerant_timestamp,
+        py_pulp_converter=lambda value: value.strftime("%Y-%m-%dT%H:%M:%SZ")
+        if isinstance(value, datetime.datetime)
+        else None,
         validator=instance_of((datetime.datetime, type(None))),
     )
     """Approximate :class:`~datetime.datetime` in UTC at which this RPM first
```

I think the problem was that the cdn_published field is declared as a datetime, but it's actually stored in pulp as a string.  Yet no converter was defined on the model for converting the datetime to a string.  The code therefore incorrectly assumed that the field should be stored in Pulp as a date and so used the $date operator while searching, which makes it unable to match.

pubtools/pulplib/_impl/model/convert.py
```diff
@@ -66,6 +66,17 @@ def tolerant_timestamp(value):
     return value
 
 
+def datetime_conversion(value):
```
rajulkumar, line 69: minor: modify the method name as "datetime_converter" or "timestamp_converter" to have a similar naming convention as other methods here.

rajulkumar (approved):

tests/client/test_search.py
```diff
@@ -100,3 +102,27 @@ def test_dict_matcher_value():
...
+        "pulp_user_metadata.cdn_published": {"$eq": "2019-09-04T00:00:00Z"}
+    }
+
+    assert filters_for_criteria(crit, FileUnit) == {
+        "pulp_user_metadata.cdn_published": {"$eq": "2019-09-04T00:00:00Z"}
+    }
+
+
+def test_cdn_published_not_datetime():
+    """criteria using cdn_published as matcher value in RpmUnit, FileUnit"""
```
rohanpm, line 122: The doc string is wrong as FileUnit hasn't been covered.

rohanpm (changes requested):
Can you please fix up the commit message and the PR title/description?  What's implemented now is completely different from what was originally, so it has to be rewritten entirely as it doesn't make any sense now.

rohanpm (approved):

PR #194, opened 2022-12-08
Use time.monotonic on py3.3+

MichalHaluza:
On Python 3.3 or newer, monotonic module aliases time.monotonic from the standard library. On older versions, it will fall back to an equivalent platform specific implementation provided by the monotonic module. This is reflected also in requirements.txt

### DIFF
--- a/.github/workflows/tox-test.yml
+++ b/.github/workflows/tox-test.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   py27:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - name: Install RPM
--- a/pubtools/pulplib/_impl/log.py
+++ b/pubtools/pulplib/_impl/log.py
@@ -1,6 +1,9 @@
 import logging
 
-from monotonic import monotonic
+try:
+    from time import monotonic
+except ImportError:  # pragma: no cover
+    from monotonic import monotonic
 
 
 class TimedLogger(object):
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ frozenlist2
 frozendict; python_version >= '3.6'
 pubtools>=0.3.0
 humanize
+monotonic; python_version < '3.3'
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@ envlist = py27,py36,static,pidiff,docs
 [testenv]
 deps=-rtest-requirements.txt
 commands=pytest -v {posargs}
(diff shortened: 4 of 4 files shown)


MichalHaluza:
@rohanpm: Is there anything I can do with the `No module named 'monotonic'` reported by pidiff? The module name is correct and the commit shouldn't be treated as an API change. Not to mention that the module has been imported and used until now, I just added it as an explicit requirement.

rohanpm:
> @rohanpm: Is there anything I can do with the `No module named 'monotonic'` reported by pidiff? The module name is correct and the commit shouldn't be treated as an API change. Not to mention that the module has been imported and used until now, I just added it as an explicit requirement.

The problem here seems to be:

- pubtools-pulplib always depended on monotonic but did not declare the dependency.
- that worked fine because more-executors *did* depend on monotonic so it was always indirectly pulled in.
- but now a new more-executors is out which doesn't depend on monotonic, so it is now possible to install pubtools-pulplib and miss installing monotonic.

In other words, the latest more-executors release caused installation of the latest pubtools-pulplib from pypi to become broken, and pidiff is just accurately reflecting that.  The broken package can reproduce at the command-line, with the pubtools-pulplib currently on pypi today:

```
$ virtualenv ~/dev/clean-venv
...

$ ~/dev/clean-venv/bin/pip install pubtools-pulplib
...

$ ~/dev/clean-venv/bin/python -c 'import pubtools.pulplib'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/rmcgover/dev/clean-venv/lib64/python3.10/site-packages/pubtools/pulplib/__init__.py", line 1, in <module>
    from ._impl.client import Client, PulpException, TaskFailedException, CopyOptions
  File "/home/rmcgover/dev/clean-venv/lib64/python3.10/site-packages/pubtools/pulplib/_impl/client/__init__.py", line 1, in <module>
    from .client import Client, CopyOptions
  File "/home/rmcgover/dev/clean-venv/lib64/python3.10/site-packages/pubtools/pulplib/_impl/client/client.py", line 26, in <module>
    from ..log import TimedLogger
  File "/home/rmcgover/dev/clean-venv/lib64/python3.10/site-packages/pubtools/pulplib/_impl/log.py", line 3, in <module>
    from monotonic import monotonic
ModuleNotFoundError: No module named 'monotonic'
```

Of course, we need to release a new pubtools-pulplib to fix this.  The most practical thing seems to be to ignore this specific CI failure.  Also if there are any tools which can verify that a library properly declares all its dependencies and doesn't rely on any of them being pulled in transitively, it would be good to adopt them...

rohanpm (approved):

PR #196, opened 2023-01-30
Drop Python2 support

lrossi-dev:
After Supercharge Pub 3 was delivered, Python 2 support is no longer required in it's dependencies as well. Dropping Python 2 will make the repos easier to maintain and will speed up Tox and Github Actions since some steps can be removed.


.github/workflows/tox-test.yml
```diff
@@ -3,20 +3,20 @@ name: Tox tests
 on: [push, pull_request]
 
 jobs:
-  py27:
-    runs-on: ubuntu-20.04
+  py36:
```
rohanpm, line 6: I don't think you need to add this because the 'coverage' job already covers python 3.

rohanpm (changes requested):

.github/workflows/tox-test.yml
```diff
@@ -3,21 +3,7 @@ name: Tox tests
...
-      - name: Setup Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: 2.7
-      - name: Install Tox
-        run: pip install tox
-      - name: Run Tox
-        run: tox -e py27
-  static:
+  tatic:
```
rohanpm, line 6: This job name has been messed up.

```suggestion
  static:
```

rohanpm (changes requested):

pubtools/pulplib/_impl/compat_attr.py
```diff
@@ -30,9 +28,7 @@ def s(*args, **kwargs):
         # don't use the attrs-generated repr by default
         kwargs["repr"] = False
 
-    if "kw_only" in kwargs and (
-        ATTR_VERSION < (18, 2) or sys.version_info < (3,)
-    ):  # pragma: no cover
+    if "kw_only" in kwargs and ATTR_VERSION < (18, 2):
         # This is only implemented for Python 3 and only in attrs 18.2 and newer.
```
rohanpm, line 32: The CI fails because the statement coverage has decreased from 100%.  I'm thinking this branch is the reason why, as it was previously marked with "pragma: no cover" (which we did for py2-only paths because the coverage config uses py3) and that pragma has been removed.

How about just removing this entire branch?  I was considering this earlier in the review anyway because I'm confident there's no reason to use such an old version of attrs on python 3.  Now that it's causing this coverage issue, we may as well go ahead and remove it.

By the way, the comment at the top of this function is also now wrong and could just say something like "set some defaults when using attr.s so they cannot be easily forgotten".

rohanpm (changes requested):

pubtools/pulplib/_impl/compat_attr.py
lrossi-dev: Can I also remove that function [fileds_dict](https://github.com/release-engineering/pubtools-pulplib/blob/master/pubtools/pulplib/_impl/compat_attr.py#L43) defined later in this same file then? It's also considering an earlier version of `attr` and might impact coverage as well.

rohanpm (approved):

PR #198, opened 2023-02-06
Repo Lock [RHELDST-14968]

amcmahon-rh:
Some pub tasks work on mutable pulp data, so if two such tasks are run at the same time one task may overwrite the results from the other. To address this, we can create a "lock" using the repository notes field. The task submits a note and waits for it to become valid, then starts executing the actual task if it's the oldest valid lock in the repository. 

First pass attempt at implementing what Rohan suggested. Looking for general comments on the overall approach. There's things like changing print statements to logs and adding tests that still need to be addressed.

I imagined this being used in a `with` statement. So the maintenance tasks become:

```
with self.pulp_client.get_repo_lock("redhat-maintenance"):
   report = self.get_maintenance_report().result()
   report = self.adjust_maintenance_report(report)
   self.set_maintenance(report).result()
```

### DIFF
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -11,6 +11,7 @@ from more_executors import Executors
 from more_executors.futures import f_map, f_flat_map, f_return, f_proxy, f_sequence
 from io import StringIO
 
+from ..model.repository.repo_lock import LOCK_CLAIM_STR
 from ..page import Page
 from ..criteria import Criteria
 from ..model import (
@@ -804,6 +805,28 @@ class Client(object):
         return self._task_executor.submit(
             self._do_request, method="POST", url=url, json=body
         )
+
+    def _get_repo_lock_data(self, repo_id):
+        repo_raw_f = self._request_executor.submit(
+            self._do_request,
+            url=os.path.join(self._url, "pulp/api/v2/repositories/%s/" % repo_id),
+            method="GET",
+        )
+        notes_f = f_map(repo_raw_f, lambda data: (data.get("notes") or {}))
+        return f_map(
+            notes_f,
+            lambda notes: {key: notes[key] for key in notes if LOCK_CLAIM_STR in key},
+        )
+
+    def _update_repo_lock_data(self, repo_id, note_delta, await_result=False):
+        update_f = self._request_executor.submit(
+            self._do_request,
+            url=os.path.join(self._url, "pulp/api/v2/repositories/%s/" % repo_id),
+            method="PUT",
+            json={"delta": {"notes": note_delta}},
+        )
+        if await_result:
+            update_f.result()
 
     def _compile_notes(self, repo):
         # Given a repo we're about to publish, calculate and set certain notes
--- a/pubtools/pulplib/_impl/fake/controller.py
+++ b/pubtools/pulplib/_impl/fake/controller.py
@@ -204,3 +204,11 @@ class FakeController(object):
         """
         with self._state.lock:
             self._state.tasks.append(task)
+
+    @property
+    def repo_lock_history(self):
+        """
+        A list containing all lock actions carried out.
+        Possible actions are: lock, unlock, multi-unlock
+        """
+        return self._state.repo_lock_history
--- a/pubtools/pulplib/_impl/model/repository/base.py
+++ b/pubtools/pulplib/_impl/model/repository/base.py
@@ -8,6 +8,7 @@ from attr import validators, asdict
 from frozenlist2 import frozenlist
 from more_executors.futures import f_proxy, f_map, f_flat_map
 
+from .repo_lock import RepoLock
 from ..attr import pulp_attrib, PULP2_FIELD, PULP2_MUTABLE
 from ..common import PulpObject, Deletable, DetachedException
 from ..convert import frozenlist_or_none_converter
@@ -594,6 +595,43 @@ class Repository(PulpObject, Deletable):
                 self.id, asdict(options, filter=lambda name, val: val is not None)
             )
         )
+
+    def lock(self, context, duration=None):
+        """
+        Obtain an exclusive advisory lock on this repository.
+
+        Returns a context manager representing the lock, intended to be used
+        via a `with` statement. When the context is entered, the caller will
+        wait until the lock can be acquired (or raise an exception if the lock
+        can't be acquired).
+
+        Only a single :class:`~pubtools.pulplib.Client` is able to hold the lock
+        on a repository at any given time. The lock does not prevent modifications
+        to the repo with the Pulp API, and does not affect other Pulp client
+        implementations or instances of :class:`~pubtools.pulplib.Client` not
+        using the `lock` method.
+
+        Args:
+            context:
+                A short description of the task being carried out with the lock.
+
+                This value will be added to the lock in the repo and may be
+                used for debugging.
+
+            duration
+                Maximum duration of the lock, in seconds.
+
+                This value is used only if this client fails to release the
+                lock (for example, because the current process is killed).
+                In this case, the duration will be used by other clients in
+                order to detect and release stale locks, avoiding a deadlock.
+
+                There is no way to extend the duration of an acquired lock,
+                so the caller should always ensure they request a `duration`
+                high enough to cover the entire expected lifetime of the lock.
+        """
+
+        return RepoLock(self.id, self._client, context, duration)
 
     def remove_content(self, criteria=None, **kwargs):
         """Remove all content of requested types from this repository.
--- a/pubtools/pulplib/_impl/model/repository/repo_lock.py
+++ b/pubtools/pulplib/_impl/model/repository/repo_lock.py
@@ -1,0 +1,221 @@
+import json
+import random
+import string
+from time import sleep
+import os
+import logging
+
+from datetime import datetime, timedelta
+from more_executors.futures import f_map
+from requests.exceptions import HTTPError
+
+LOCK_CLAIM_STR = "pulplib-lock-claim-"
+
+LOCK_VALID_FROM_OFFSET = int(os.getenv("PULPLIB_LOCK_VALID_FROM_OFFSET", "10"))
+LOCK_EXPIRATION_OFFSET = int(os.getenv("PULPLIB_LOCK_EXPIRATION_OFFSET", "1800"))
+LOCK_EXPIRATION_OFFSET_MIN = int(os.getenv("PULPLIB_LOCK_EXPIRATION_OFFSET_MIN", "300"))
+LOCK_EXPIRATION_OFFSET_MAX = int(
+    os.getenv("PULPLIB_LOCK_EXPIRATION_OFFSET_MAX", "216000")
+)
+
+AWAIT_ACTIVATION_SLEEP = int(os.getenv("PULPLIB_AWAIT_ACTIVATION_SLEEP", " 10"))
+
+LOG = logging.getLogger("pubtools.pulplib")
+
+
+# datetime is a built-in so cannot be mocked directly.
+# This function is a workaround to allow mocking for testing.
+def now():
+    return datetime.now()  # pragma: no cover
+
+
+class LockClaim(object):
+    def __init__(self, **kwargs):
+        expiration_offset = kwargs.get("expiration_offset")
+        expiration_offset = (
+            expiration_offset if expiration_offset else LOCK_EXPIRATION_OFFSET
+        )
+        expiration_offset = max(
+            min(expiration_offset, LOCK_EXPIRATION_OFFSET_MAX),
+            LOCK_EXPIRATION_OFFSET_MIN,
+        )
+        self.id = (
+            kwargs["id"]
+            if "id" in kwargs
+            else "".join(random.choices(string.ascii_letters + string.digits, k=8))
+        )
+        self.context = kwargs["context"]
+        self.created = (
+            datetime.fromtimestamp(kwargs["created"]) if "created" in kwargs else now()
+        )
+        self.valid_from = (
+            datetime.fromtimestamp(kwargs["valid_from"])
+            if "valid_from" in kwargs
+            else self.created + timedelta(seconds=LOCK_VALID_FROM_OFFSET)
+        )
+        self.expires = (
+            datetime.fromtimestamp(kwargs["expires"])
+            if "expires" in kwargs
+            else self.created + timedelta(seconds=expiration_offset)
+        )
+
+    @property
+    def as_json(self):
+        data = {
+            "id": self.id,
+            "context": self.context,
+            "created": datetime.timestamp(self.created),
+            "valid_from": datetime.timestamp(self.valid_from),
+            "expires": datetime.timestamp(self.expires),
+        }
+        return json.dumps(data)
+
+    def __lt__(self, other):
+        return self.created < other.created
+
+    def __eq__(self, other):
+        return self.id == other.id
+
+    @staticmethod
+    def from_json_data(json_data):
+        return LockClaim(**json_data)
+
+    @property
+    def is_valid(self):
+        return self.valid_from < now() < self.expires
+
+    @property
+    def is_expired(self):
+        return now() > self.expires
+
+
+class RepoLock(object):
+    def __init__(
+        self,
+        repo_id,
+        client,
+        lock_context,
+        expiration_offset,
+    ):
+        self._repo_id = repo_id
+        self._client = client
+        self._lock_claim = None
+        self._lock_context = lock_context
+        self._expiration_offset = expiration_offset
+
+    def __enter__(self):
+        LOG.info("Attempting to lock %s.", self._repo_id)
+        self.remove_expired_locks()
+        self._lock_claim = LockClaim(
+            context=self._lock_context, expiration_offset=self._expiration_offset
+        )
+        self.submit_lock_claim()
+        self.await_lock_activation()
+
+    def __exit__(self, exc_type, exc_value, tb):
+        if self._lock_claim.is_expired:
+            logging.error(
+                "Client requested lock on repo %s for %s seconds "
+                "but was held for %s seconds (%s)",
+                self._repo_id,
+                (self._lock_claim.expires - self._lock_claim.created).seconds,
+                (now() - self._lock_claim.created).seconds,
+                self._lock_claim.context,
+            )
+        self.delete_lock_claim()
+        self._lock_claim = None
+
+    def get_repo_lock_data(self, filter_invalid=False):
+        LOG.debug("Getting current repo lock data for %s.", self._repo_id)
+        locks_f = self._client._get_repo_lock_data(self._repo_id)
+        locks_f = f_map(
+            locks_f,
+            lambda locks: [
+                LockClaim.from_json_data(json.loads(l)) for l in locks.values()
+            ],
+        )
+        parsed_locks = locks_f.result()
+        parsed_locks.sort()
+        if filter_invalid:
+            parsed_locks = [l for l in parsed_locks if l.is_valid]
+        return parsed_locks
+
+    def await_lock_activation(self):
+        # if `now` is later than `valid_from`, the timedelta is negative and
+        # the seconds value becomes a whole day minus the timedelta ~ 86400.
+        # total_seconds returns a negative time value, which we clamp.
+        sleep(max((self._lock_claim.valid_from - now()).total_seconds(), 0))
+        while True:
+            lock_claims = self.get_repo_lock_data(filter_invalid=True)
+            # If our claim is the oldest valid lock claim.
+            if self._lock_claim == lock_claims[0]:
+                LOG.info("Lock obtained for %s.", self._repo_id)
+                return
+            LOG.info("Awaiting lock for %s.", self._repo_id)
+            sleep(AWAIT_ACTIVATION_SLEEP)
+
+    def submit_lock_claim(self):
+        LOG.info(
+            "Submitting lock with id '%s' to %s (%s).",
+            self._lock_claim.id,
+            self._repo_id,
+            self._lock_claim.context,
+        )
+        self._client._update_repo_lock_data(
+            self._repo_id,
+            {LOCK_CLAIM_STR + self._lock_claim.id: self._lock_claim.as_json},
+        )
+
+    def delete_lock_claim(self):
+        LOG.info(
+            "Deleting lock with id '%s' from %s (%s).",
+            self._lock_claim.id,
+            self._repo_id,
+            self._lock_claim.context,
+        )
+        note_delta = {LOCK_CLAIM_STR + self._lock_claim.id: None}
+        try:
+            self._client._update_repo_lock_data(
+                self._repo_id, note_delta, await_result=True
+            )
+        except HTTPError as e:
+            if self._pulp_exception_from_missing_lock(e, list(note_delta.keys())):
+                logging.error("Lock %s has already been deleted.", self._lock_claim.id)
+            else:
+                raise e  # pragma: no cover
+
+    def _pulp_exception_from_missing_lock(self, e, lock_names):
+        # If we try to delete a lock from pulp notes that's already been
+        # deleted, it will result in a 500 error due to a KeyError exception
+        # If multiple locks are sent for deletion, it appears Pulp will fail on
+        # the first missing one and not carry out any deletions.
+        lock_names = ["KeyError: '%s'\n" % lock for lock in lock_names]
+        return any(
+            [exception in lock_names for exception in e.response.json()["exception"]]
+        )
+
+    def remove_expired_locks(self):
+        expired_locks = [lock for lock in self.get_repo_lock_data() if lock.is_expired]
+        if expired_locks:
+            for lock in expired_locks:
+                LOG.warning(
+                    "Removing expired lock with id %s from %s (%s).",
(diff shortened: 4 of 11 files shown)


amcmahon-rh:
Last Friday I tested this out with a small script that added/ removed records from the QA "redhat-maintenance" repository. This worked as expected and checking `repos.json` showed the modifications from all the scripts. 


Here's a [Screenshot](https://drive.google.com/file/d/1PZ9VEuSuYoY2gKY8X18oHnUHbaTWMQQd/view?usp=sharing) of the scripts in progress. I had all the commands written out and started each script within a second of each other. One thing that might stand out is on the bottom right it jumps from 3rd to 1st in the queue. That's because the sleep between querying the notes is based on how many locks there are in the queue. While the lock was in 3rd place, the sleep delay was long enough to skip over a query while it was 2nd place.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -448,6 +448,14 @@ def update_repository(self, repository):
 
         return out
 
+    def get_repo_lock(self, repo_id):
```
rohanpm, line 451: I think this might be better off living on the Repository class, and also with no 'get' prefix, so the caller would write something like:

```python
with repo.lock():
    repo.do_some_stuff(...)
```

Whether it lives on Client or on Repository is somewhat arbitrary, either will technically work.  I feel that Repository is more consistent with the rest of the API which exists today.

Also, it forces the caller to verify the repo's existence first. That might seem a negative (as you're forced to do one extra arguably unnecessary request), but in fact in many places where we first tried to make use of a repo without explicitly verifying it exists first, users had trouble understand the error messages if they passed the ID of a nonexistent repo.

It would be good also if we don't document the object returned by lock() and don't make it public API.  i.e. we document that it can be used as a context manager, and that's it.  This avoids us getting locked into an implementation that we might not be able to support long-term.

pubtools/pulplib/_impl/client/repo_lock.py
```diff
@@ -0,0 +1,161 @@
+import json
+import random
+import string
+from time import sleep
+
+from datetime import datetime, timedelta
+from more_executors.futures import f_map
+
+
+LOCK_CLAIM_STR = "lock-claim-"
```
rohanpm, line 10: It should probably start with "pulplib-" so it's possible to trace these back to the library which created them, and also avoid clashes with anything else.

pubtools/pulplib/_impl/client/repo_lock.py
```diff
@@ -0,0 +1,161 @@
...
+
+LOCK_VALID_FROM_OFFSET = 10
+LOCK_EXPIRATION_OFFSET = 3600
```
rohanpm, line 13: I'm pretty sure that the expiration should be part of the public API.  This could be a default, but only the caller has an idea of how long they'd expect to be holding the lock, so it'd be best if they can set a smaller expiration in some cases.

For example in the maintenance-on/off case, it should only need the lock for a few seconds.  The default would be safe but it would also be overkill.

The other magic numbers here probably don't need to be public API though we generally make them tunable by "hidden" environment variables, so that we can at least tweak them in emergency scenarios without code changes if that turns out to be needed.

pubtools/pulplib/_impl/client/repo_lock.py
```diff
@@ -0,0 +1,161 @@
...
+
+# How long to sleep before checking if we can activate the lock again
+# This should be a sleep per locks ahead of us. If there's 3 locks ahead of
```
rohanpm, line 16: > This should be a sleep per locks ahead of us

I feel like this part of the logic is unnecessary and might be better off being dropped.
I guess the benefit is only to reduce the number of requests to the server, but it seems negligible.

On the downside seems likely to introduce unnecessary delays.  I'd expect in practice that clients often need to hold the lock for a fair bit less than 15 seconds, so this could easily introduce a half minute delay where it's just sleeping even though everyone else already released their locks.

pubtools/pulplib/_impl/client/repo_lock.py
```diff
@@ -0,0 +1,161 @@
...
+        self._lock_claim = LockClaim()
+        self.submit_lock_claim()
+        self.await_lock_activation()
+
+    def __exit__(self, exc_type, exc_value, tb):
+        self.delete_lock_claim()
+        self._lock_claim = None
+
+    def get_repo_lock_data(self, filter_invalid=False):
+        print("Getting current repo lock data")
```
rohanpm, line 89: When you're done with these, rather than removing the prints you should probably change them to debug or info logs.

pubtools/pulplib/_impl/client/repo_lock.py
```diff
@@ -0,0 +1,161 @@
...
+                try:
+                    print(
+                        "Lock not yet valid: position in queue "
+                        + str(lock_claims.index(self._lock_claim))
+                    )
+                    sleep(
+                        AWAIT_LOCK_ACTIVATION_SLEEP
+                        * lock_claims.index(self._lock_claim)
+                    )
+                except ValueError:
```
rohanpm, line 130: I'm not sure that we really need to catch this at all.  This would happen if, for example, the lock notes contain corrupt data?

You can't really do anything other than raise an exception in such cases, and if you raise something other than *the original* exception it arguably just obscures the error.  Also, if the backtrace shows more than one exception, a lot of developers will not read it properly and will randomly pick whichever of the exceptions they noticed first as being the root cause...

rohanpm (commented):
Looks pretty good to me.  I know you were only looking for high-level comments but I put inline comments anyway to save time.

May I suggest to you also that your script you've been using for testing could be turned into a proper example (see `examples` directory).  A lot of the library features have an example script in the repo which exercises that one feature independently and also gives you a way to enable loggers at DEBUG level.

If you do that first, you might find this more efficient to develop, e.g. you won't need to go back and delete 'print' statements later since you can use proper logging right from the beginning (and whatever logs you need while initially developing this, might be needed later in debugging it, so it's potentially better to implement them as DEBUG logs than prints anyway).

amcmahon-rh:
@rohanpm 

I've gone through and addressed your comments :)

I'm not entirely sure how I feel about having the AWAIT_ACTIVATION_SLEEP and LOCK_EXPIRATION_OFFSET values in the repo base file. I played with passing a kwarg dict instead but it ""looked yucky"". Though it's probably also a case of fixating too much on specific implementation details.

pubtools/pulplib/_impl/model/repository/repo_lock.py
```diff
@@ -0,0 +1,163 @@
...
+from time import sleep
+import os
+import logging
+
+from datetime import datetime, timedelta
+from more_executors.futures import f_map
+
+LOCK_CLAIM_STR = "pulplib-lock-claim-"
+
+LOCK_VALID_FROM_OFFSET = os.getenv("PULPLIB_LOCK_VALID_FROM_OFFSET", 10)
```
rohanpm, line 13: I don't think these actually work right now if you set the env vars.  There'll be a type error.

If you consider the expression:

```python
os.getenv("PULPLIB_LOCK_VALID_FROM_OFFSET", 10)
```

- if the caller didn't set the env var, it returns an int.
- if the caller set the env var, it returns a str (because env vars are always strings).

The code using these values unconditionally tries to use them as integers (e.g. using with 'timedelta' class).  So if you try to set one of these env vars and then run the example, I'm betting it'll crash.

Given that getenv will return a str in the 'env var is not set' case, it's best to make the expression *also* return a str in the 'env var is set' case, that way you can't forget to convert it to an int.

I usually prefer this pattern instead:

```suggestion
LOCK_VALID_FROM_OFFSET = int(os.getenv("PULPLIB_LOCK_VALID_FROM_OFFSET") or "10"))
```

...which will do the conversion properly in both set and unset cases; and also if the caller has set PULPLIB_LOCK_VALID_FROM_OFFSET to an empty string, this is handled as "use the default", rather than "crash because empty string can't be converted to int".

pubtools/pulplib/_impl/model/repository/repo_lock.py
```diff
@@ -0,0 +1,163 @@
...
+        activation_await_sleep=AWAIT_ACTIVATION_SLEEP,
+    ):
+        self._repo_id = repo_id
+        self._request_func = request_func
+        self._lock_claim = None
+        self._expiration_offset = expiration_offset
+        self._activation_await_sleep = activation_await_sleep
+
+    def __enter__(self):
+        LOG.info("Attempting to lock %s." % self._repo_id)
```
rohanpm, line 86: Can you please let the logging framework do the string formatting instead, like this:

```suggestion
        LOG.info("Attempting to lock %s.", self._repo_id)
```

This is actually a pylint complaint that we should usually obey - https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-format-interpolation.html. Many legacy projects ignore it but we should try to stick with it in the newer stuff.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -595,6 +597,30 @@ def sync(self, options=None):
...
+                seconds.
+        """
+        return RepoLock(
+            self.id,
+            lambda url, **kwargs: self._client._request_executor.submit(
+                self._client._do_request,
+                url=os.path.join(self._client._url, url),
+                **kwargs
+            ),
+            expiration_offset=duration,
```
rohanpm, line 620: I'd like to request that we clamp the caller's input to a reasonable range somewhere in here, because of the high impact in case the user provides bad values.  e.g.

- if the caller passes a too low duration, we can end up with the same silent data corruption we're trying to fix here.
- if the caller passes a too high duration, we can end up with extreme delays/apparent deadlock in the cases where a stale lock needs to be released.

I'll make a rough guess:

- if caller requested a duration less than 5 minutes, we should force it to 5 minutes (because basically anything could take that long if there's an outage which forces requests to be retried).
- if caller requested a duration more than 6 hours, we should force it to 6 hours (kind of arbitrarily picked as one quarter of a day and fitting into one working day... and nothing comes to mind which should ever genuinely need the lock for longer than that.)

I prefer "clamp" rather than "validate" because if the approach is to validate, then later changes to it are potentially backwards-incompatible (e.g. if we accept 5 minutes today but later need to change it to 15 minutes, that could break someone).

pubtools/pulplib/_impl/model/repository/repo_lock.py
```diff
@@ -0,0 +1,163 @@
...
+        self.remove_expired_locks()
+        self._lock_claim = LockClaim(expiration_offset=self._expiration_offset)
+        self.submit_lock_claim()
+        self.await_lock_activation()
+
+    def __exit__(self, exc_type, exc_value, tb):
+        self.delete_lock_claim()
```
rohanpm, line 93: While I wouldn't call it absolutely necessary, there is a bit of a problem right now in this scenario:

- caller requests a lock of duration X
- the time between enter/exit ends up being *longer than* X

i.e. the caller requested a too low duration.  This means the caller didn't actually hold the lock for as long as they needed it.

While this is the caller's fault for requesting a too low duration, and we can't prevent it, maybe we should at least detect it?  e.g. at exit time, should we do one of the following?

1. check if current time is less than the expiry time; or...
2. fetch the current locks and verify that our lock still exists (before deleting it)

...and then log a warning?  Otherwise, in cases where a client requests too small duration, there'll be no trace of this happening - just silent data corruption, the same kind of corruption we're trying to fix.

Option (1) seems pretty easy to implement so maybe it's worthwhile.  e.g. it could generate a message along the lines of:

`ERROR: client requested Pulp repo lock for $X seconds but held lock for $Y seconds!`

It's definitely a bug so ERROR seems right, though not sure it's severe enough to raise.

pubtools/pulplib/_impl/model/repository/repo_lock.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+
+    def remove_expired_locks(self):
+        invalid_locks_delta = {
+            LOCK_CLAIM_STR + lock.id: None
+            for lock in self.get_repo_lock_data()
+            if not lock.is_valid
+        }
+        if invalid_locks_delta:
+            LOG.debug("Removing invalid locks from %s." % self._repo_id)
```
rohanpm, line 155: Given that this only happens if someone has crashed, I think this is worthy of WARNING log level, and probably mentioning every affected lock ID as well.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -595,6 +597,30 @@ def sync(self, options=None):
...
+        just provides a way to pause tasks to prevent race conditions.
+
+        Args:
+            duration
+                How long the lock should remain valid for in seconds.
+            retry_wait:
+                How long to wait between attempting to activate the lock in
+                seconds.
+        """
+        return RepoLock(
```
rohanpm, line 613: Sorry I haven't realized this earlier, but I think we ought to have a kind of 'name' or 'context' string field as well which is stored on the lock.  It's just for logging and debugging purposes.

The string would be provided by the caller and should briefly describe why you have the lock, for example:

```python
with repo.lock('updating maintenance report'):
    # ...the steps to fetch, modify and re-upload maintenance report
```

It doesn't have to be unique across all existing locks, but ideally is unique for each code path obtaining a lock.

This wouldn't affect the locking mechanism, the point of it is just to make it easier to trace who obtained a lock. 

e.g. imagine we're facing a problem that some tasks are running slowly because there often seems to be some stale locks on a repo.  We need to figure out who is creating those locks and failing to release them.

If the locks included a field like `{"context": "updating maintenance report"}`, we could simply grep for `updating maintenance report` across our code and immediately find the code we should start debugging.  Otherwise we would need to check every usage of .lock(), cross-reference lock IDs with logs in Splunk (hoping all relevant log messages made it there), and so on.

Also, the log messages for acquiring/releasing/deleting stale locks could include this string in the message.  That'll be useful in some cases, e.g. when we delete a stale lock, if we generated a message like:

```
WARNING: removing expired lock pulplib-lock-claim-abc123 (updating maintenance report)
```

Then anyone reading the log could immediately know that removing this expired lock might affect other tasks relating to "updating maintenance report" and shouldn't affect anything else.

rohanpm (changes requested):

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -597,18 +597,38 @@ def sync(self, options=None):
...
+
+                This value will be added to the lock in the repo and may be
+                used for debugging.
+
             duration
-                How long the lock should remain valid for in seconds.
-            retry_wait:
-                How long to wait between attempting to activate the lock in
-                seconds.
+                 Maximum duration of the lock, in seconds.
```
rohanpm, line 622: Seems like an off-by-one whitespace error here, this will probably mess with the docs rendering (should be possible to check by `tox -e docs`).

```suggestion
                Maximum duration of the lock, in seconds.
```

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -595,6 +597,50 @@ def sync(self, options=None):
...
+        via a `with` statement. When the context is entered, the caller will
+        wait until the lock can be acquired (or raise an exception if the lock
+        can't be acquired).
+
+        Only a single :class:`~pubtools.pulplib.Client` is able to hold the lock
+        on a repository at any given time. The lock does not prevent modifications
+        to the repo with the Pulp API, and does not affect other Pulp client
+        implementations or instances of :class:`~pubtools.pulplib.Client` not
+        using the `lock` method.
+        Args:
```
rohanpm, line 614: Needs a blank line before it:

```suggestion

        Args:
```

pubtools/pulplib/_impl/model/repository/repo_lock.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+# datetime is a built-in so cannot be mocked directly.
+# This function is a workaround to allow mocking for testing.
+def now():
+    return datetime.now()
+
+
+def timestamp(date):
+    # datetime.timestamp was added in py 3.3
```
rohanpm, line 34: There's no need to support any python other than 3.3.

rohanpm: ...meant to say "older than 3.3" of course.

pubtools/pulplib/_impl/model/repository/repo_lock.py
rohanpm: Wow, it seems like pylint these days detects the same issue, which I didn't know about.  The `static` CI config currently complains about this:

```
pubtools/pulplib/_impl/model/repository/repo_lock.py:14:25: W1508: os.getenv default type is builtins.int. Expected str or None. (invalid-envvar-default)
```

examples/repo_lock.py
```diff
@@ -0,0 +1,74 @@
+from pubtools.pulplib import Client
```
rohanpm, line 1: Can you please fix up some consistency issues between this example and the others?

- naming: `_` vs `-`
- naming: `.py` vs no extension
- should be marked executable and start with a shebang line

rohanpm (changes requested):
I didn't think to check it earlier, but I've checked now and I see the new feature doesn't work at all when using the fake client ( https://release-engineering.github.io/pubtools-pulplib/api/testing.html ).

It'll crash like this:

```
  File "/home/rmcgover/src/pubtools-pulplib/pubtools/pulplib/_impl/model/repository/base.py", line 636, in <lambda>
    lambda url, **kwargs: self._client._request_executor.submit(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'FakeClient' object has no attribute '_request_executor'
```

Can you please make it work also for the fake client?

The bare minimum of "make it work" in this case means just "doesn't crash", I wouldn't insist on implementing any actual locking, though it could be done.

examples/repo_lock.py
amcmahon-rh: Hey Rohan, do you have any general advice for how to "make it work" regarding the FakeClient? 

The request function lambda requires three attributes that don't exist in the FakeClient class. A compounding issue is that one of these attributes then tries to get the method. Is this just a case of yes everything there will need to be set up? 

Looking through the repository code I can't find any special considerations for the FakeClient class, so I assume it's a case of modifying the FakeClient class to get it to work with the rest of the code as is.

rohanpm: Firstly, though I said just "don't crash" could be the goal before, later I thought ahead to the changes needed in pubtools-pulp-maintenance-on to make use of this new feature and how it would be covered in unit tests there.

In fact it would be quite valuable if there were a way, when using the fake client, to access the repo lock history (same as today you can access the repo publish history).  The way of testing the pubtools-pulp-maintenance-on change would then be as easy as, in existing test_maintenance_on test there, something like:

```
# It should have locked and then unlocked the repo
assert len(controller.repository_lock_history) == 2
assert controller.repository_lock_history[0].repository.id == "redhat-maintenance"
assert controller.repository_lock_history[0].event_type == "lock"
assert controller.repository_lock_history[1].repository.id == "redhat-maintenance"
assert controller.repository_lock_history[1].event_type == "unlock"
```

So I think making something like the above work should be the goal.

The usual approach so far has been for the real and fake clients to implement some of the same (private) API.  For example, they both implement a `def _publish_repository` which takes care of real or fake publishing.  A fair bit of refactoring would be needed here to follow that approach though.

I wouldn't expect this to be handled at the HTTP request layer, i.e. adding `_request_executor` and `_do_request` onto the fake client wouldn't be a good approach I think.

I think there are two reasonable approaches to take here:

1) refactor it so it works similarly to the current approach already used for repo publishing (`def _publish_repository` and file uploads (`_do_upload`). Give real and fake clients different implementations of a compatible API for locking/unlocking repos, then use it from `RepoLock` class.  Real client does the same kind of HTTP requests as currently prepared in `RepoLock`, fake client always succeeds (as long as repo exists) and just stores a lock/unlock event on a new attribute in FakeState class.

2) refactor so that `RepoLock` receives a reference to a client object.  Then it can explicitly check if the client object is a fake client and, if so, it can jump to a different simplified implementation.

tests/repository/test_repo_lock.py
```diff
@@ -0,0 +1,164 @@
...
+            "pulplib-lock-claim-1234": '{"id": "1234", "context": "test", "created": 1000.0, "valid_from": 1010.0, "expires": 2800.0}'
+        },
+        {"pulplib-lock-claim-1234": None},
+    ]
+
+    # Runs __enter__ and __exit__ in RepoLock
+    with lock:
+        pass
+
+    assert ctrl._state.repo_lock_history == expected_calls
```
rohanpm, line 103: A couple of issues with this:

- the `_state` property isn't public API, so there still isn't a proper way to use this from other projects.  Properties on `_state` which are meant to be public can be accompanied by an accessor on FakeController class, which is also where you'd write a doc string for it.
- the format is nasty - string values.  Is the user supposed to parse them as JSON?  We should avoid exposing the implementation detail that locks are achieved by updating repo notes which happens to use JSON documents.

Ideally this would be consistent with publish_history, sync_history and return named tuples.

pubtools/pulplib/_impl/model/repository/repo_lock.py
```diff
@@ -0,0 +1,222 @@
...
+LOCK_CLAIM_STR = "pulplib-lock-claim-"
+
+LOCK_VALID_FROM_OFFSET = int(os.getenv("PULPLIB_LOCK_VALID_FROM_OFFSET", "10"))
+LOCK_EXPIRATION_OFFSET = int(os.getenv("PULPLIB_LOCK_EXPIRATION_OFFSET", "1800"))
+LOCK_EXPIRATION_OFFSET_MIN = int(os.getenv("PULPLIB_LOCK_EXPIRATION_OFFSET_MIN", "300"))
+LOCK_EXPIRATION_OFFSET_MAX = int(
+    os.getenv("PULPLIB_LOCK_EXPIRATION_OFFSET_MAX", "216000")
+)
+
+AWAIT_ACTIVATION_SLEEP = os.getenv("PULPLIB_AWAIT_ACTIVATION_SLEEP", 10)
```
rohanpm, line 21: This still has the same bug as above used to (it will return an int if the env var is not set, and a str if the env var is set).  I suppose if you set the env var, the later sleep(AWAIT_ACTIVATION_SLEEP) will crash.

pubtools/pulplib/_impl/fake/controller.py
```diff
@@ -204,3 +204,12 @@ def insert_task(self, task):
         """
         with self._state.lock:
             self._state.tasks.append(task)
+
+    @property
+    def repo_lock_history(self):
+        """
+        A list containing all lock actions carried out.
+        Each entry is a list itself, since several lock updates can be made
```
rohanpm, line 212: In which case does this happen?  Is it only when the client decides to remove expired locks and there's more than one at a time?

It seems like a bit of a niche case, while having it be a list-of-list is more cumbersome to use than just a list, so I'm wondering if it'd be better to just flatten it.

By "more cumbersome to use", I mean for example where it has already been used in the tests:

```python
    assert len(lock_history) == 2
    assert lock_history[0][0].repository == "test-repo"
    assert lock_history[0][0].action == "lock"
    assert lock_history[1][0].repository == "test-repo"
    assert lock_history[1][0].action == "unlock"
```

This test doesn't actually verify that there were two lock-related operations.  It verifies that there were two *groups of* lock operations and it's never verified that each group contains only the expected operation (e.g. there's no `assert len(lock_history[0]) == 1`.  So that test can pass even if there are extra unexpected lock operations, and you've got to check at multiple levels of the structure if you want to avoid that.  Seems annoying to use.

pubtools/pulplib/_impl/model/repository/repo_lock.py
```diff
@@ -0,0 +1,222 @@
...
+LOCK_CLAIM_STR = "pulplib-lock-claim-"
+
+LOCK_VALID_FROM_OFFSET = int(os.getenv("PULPLIB_LOCK_VALID_FROM_OFFSET", "10"))
+LOCK_EXPIRATION_OFFSET = int(os.getenv("PULPLIB_LOCK_EXPIRATION_OFFSET", "1800"))
+LOCK_EXPIRATION_OFFSET_MIN = int(os.getenv("PULPLIB_LOCK_EXPIRATION_OFFSET_MIN", "300"))
+LOCK_EXPIRATION_OFFSET_MAX = int(
+    os.getenv("PULPLIB_LOCK_EXPIRATION_OFFSET_MAX", "216000")
+)
+
+AWAIT_ACTIVATION_SLEEP = int(os.getenv("PULPLIB_AWAIT_ACTIVATION_SLEEP", 10))
```
rohanpm, line 21: This will technically work now, though the second parameter to getenv being an int rather than str is inconsistent with all the previous lines for no reason.  Not sure if pylint will still complain about this too.

rohanpm (commented):
Looks pretty good, do you want to do final steps on this now?

- remove WIP from the pull request message
- ideally squash all the commits into just one
- rebase it to fix the conflict and get a fresh CI run

rohanpm:
> The bandit test is failing with request_without_timeout for every detected issue. We discussed this earlier in the week since it was affecting rcm-pub, and there's already a bug lodged with bandit for false positives. [PyCQA/bandit#996](https://github.com/PyCQA/bandit/issues/996)

There is also RHELDST-16791 which says that we should be pinning the bandit version here, like we do with other packages used in CI, so the CI will not flip from passing to failing when nobody made changes breaking anything.

PR #199, opened 2023-02-07
Added new field container_list to Erratum model

midnightercz:


pubtools/pulplib/_impl/model/unit/erratum.py
```diff
@@ -374,6 +374,15 @@ class ErratumUnit(Unit):
...
     """A list of package collections associated with the advisory."""
 
+    container_list = pulp_attrib(
+        type=list,
+        pulp_field="pulp_user_metadata.container_list",
+        converter=frozenlist_or_none_converter,
+        default=None,
+        validator=optional_list_of(str),
+    )
+    """A list of package collections associated with the advisory."""
```
rohanpm, line 384: Doc string is wrong.  Could it please be updated for accuracy?

It should mention the format of the strings or at least give an example or two.

pubtools/pulplib/_impl/model/unit/erratum.py
```diff
@@ -374,6 +374,15 @@ class ErratumUnit(Unit):
     )
     """A list of package collections associated with the advisory."""
 
+    container_list = pulp_attrib(
+        type=list,
+        pulp_field="pulp_user_metadata.container_list",
```
rohanpm, line 379: Can you please double-check if a flat list of string is really sufficient?

This comment https://url.corp.redhat.com/952c018 said that it should also contain arch info.

rohanpm (changes requested):

pubtools/pulplib/_impl/model/unit/erratum.py
midnightercz: yeah, it's actually slightly complicated structure

pubtools/pulplib/_impl/model/validate.py
```diff
@@ -26,5 +26,82 @@ def __call__(self, inst, attr, value):
...
+
+    def __repr__(self):
+        return "<named_mapping_validator {mapping!r}>".format(mapping=self.mapping)
+
+
+def named_mapping_validator(mapping):
+    return NamedMappingValidator(mapping)
+
+
+@s(kw_only=True, frozen=True, slots=True)
```
rohanpm, line 69: Hmm, I don't think the usage of attr.s does much, does it?

pubtools/pulplib/_impl/model/validate.py
```diff
@@ -26,5 +26,82 @@ def __call__(self, inst, attr, value):
...
+class ContainerListValidator(object):
+    def __call__(self, inst, attr, value):
+        print(value)
```
rohanpm, line 72: Please drop the print.

pubtools/pulplib/_impl/model/unit/erratum.py
```diff
@@ -374,6 +380,15 @@ class ErratumUnit(Unit):
...
+        converter=frozenlist_or_none_converter,
```
rohanpm, line 386: There is a problem here, if the field contains a list of dict, it will be mutable.  frozenlist will be applied only at the object's top level, dicts within the list will be mutable.  All the fields need to be deeply immutable.

If you want it to be a complex structure with lists containing dicts containing lists and so on, please make sure there's a converter to make it all deeply immutable, i.e. all lists replaced with frozenlist and all dicts replaced with frozendict.

The existing frozendict_or_none_converter almost does this already, maybe some of that can be reused.

tests/data/sample-erratum.json
```diff
@@ -701,6 +701,24 @@
...
         "content_types": [
             "module"
+        ],
+        "container_list": [
+            {
+              "e2e/test-operator-container": {
+                "digest": "sha256:2321a7d13d9fa53f05437663cf2dc217d15f3cda4b67076c941b10f0491cf9d7",
+                "images": {
+                  "docker-image-sha256:06d1c5e4fa6a5d1ff868388f3feadf193d04128b62d1181e37fe4ab8ecda27e1.s390x.tar.gz": {
+                    "digest": null
```
rohanpm, line 711: What's the digest expected to be in practice?  Can we please get at least one case where it's non-null?

pubtools/pulplib/_impl/model/validate.py
```diff
@@ -26,5 +26,82 @@ def __call__(self, inst, attr, value):
...
+    def __repr__(self):
+        return "<named_mapping_validator {mapping!r}>".format(mapping=self.mapping)
+
+
+def named_mapping_validator(mapping):
+    return NamedMappingValidator(mapping)
+
+
+@s(kw_only=True, frozen=True, slots=True)
+class ContainerListValidator(object):
```
rohanpm, line 70: As the validator is fairly complicated, could we please get at least one test which tries to construct an object with *invalid* input and verifies that this fails?

rohanpm (changes requested):

pubtools/pulplib/_impl/model/validate.py
midnightercz: I took this from attrs source where it was on another validator. I guess it's just authors obsession by attrs, but on the other hand you don't need to provide there __init__

rohanpm: Well, it is not harmful either and I guess at least is going to implement a default `repr` and so on, so maybe that's why.  Does not seem to matter whether it's left here or not.

tests/unit/test_sample_erratum.py
```diff
@@ -689,6 +692,38 @@ def test_can_load_erratum(data_path):
         references=[],
         pkglist=[],
         content_type_id="erratum",
+        container_list=frozenlist(
```
rohanpm, line 695: You shouldn't have to explicitly put the frozenlist/frozendict here and it'd probably be better if you didn't - there's supposed to be an automatic conversion anyway when the object is initialized, and I think that's how it would generally be used.  The usage of frozenlist/frozendict is somewhat hidden (caller doesn't care about it, as long as they never try to mutate the data).

tests/data/sample-erratum.json
```diff
@@ -701,6 +701,24 @@
     "pulp_user_metadata": {
         "content_types": [
             "module"
+        ],
+        "container_list": [
+            {
+              "e2e/test-operator-container": {
+                "digest": "sha256:2321a7d13d9fa53f05437663cf2dc217d15f3cda4b67076c941b10f0491cf9d7",
+                "images": {
+                  "docker-image-sha256:06d1c5e4fa6a5d1ff868388f3feadf193d04128b62d1181e37fe4ab8ecda27e1.s390x.tar.gz": {
```
rohanpm, line 710: I think this needs to be on hold until thread https://github.com/release-engineering/pushsource/pull/305/files#r1105144319 in pushsource is settled.

We can keep the discussion there mainly, but I would expect the outcome to flow on to here as well.

This is what the data we expose from Pulp is going to look like, and I'm pretty sure that the people asking for this to be added to Pulp care about the arch and do not care about the koji archive name.

rohanpm (changes requested):

amcmahon-rh:
For some reason the test I wrote passes locally but fails with github actions. I'll be doing a few more pushes to try debug this.

setup.py
```diff
@@ -21,7 +21,7 @@ def get_requirements():
 
 setup(
     name="pubtools-pulplib",
-    version="2.34.5",
+    version="2.36.0",
```
rohanpm, line 24: Seems like this is being bumped more than necessary, perhaps intentional because there is another PR in progress which could potentially be released as 2.35.0 prior to this one?  Anyway, this should not be harmful, but if they are really submitted around the same time then one of the PRs is doomed to conflict.

rohanpm (approved):

setup.py
midnightercz: This version was suggested by pidiff I think even on my local machine

rohanpm (approved):

PR #200, opened 2023-02-27
Fixed container_list tests

midnightercz:


tests/model/test_errata_model.py
```diff
@@ -1,19 +1,29 @@
...
 import pytest
 
-from pubtools.pulplib import ErratumUnit
+from pubtools.pulplib import ErratumUnit, Unit
 
 from pubtools.pulplib import InvalidDataException
 
 
+def test_container_list(data_path):
+    """Provide invalid structure object to container_list attribute."""
```
rohanpm, line 12: That's not what the test is doing, is it?  It doesn't seem like the test data file named `sample-erratum.json` should contain invalid data.

rohanpm (changes requested):

tests/model/test_errata_model.py
midnightercz: True, I copy pasted it from the existing one

rohanpm: With the latest change I think you updated the doc string on the wrong test.  Now each test's doc string says the *opposite* of what it's testing...

rohanpm (changes requested):

rohanpm (approved):

rajulkumar (approved):

PR #203, opened 2023-04-03
Revert "Remove workarounds for old version of attrs library"

MichalHaluza:
This reverts commit 73e2568da443dea47c79149e64762311a1acbf01.

The workarounds are still needed on RHEL 8 where the python 3.6 libraries are installed from RPMs as RHEL 8 ships attrs 17.4


MichalHaluza:
I'm not addressing the bandit complaints about mocked requests timeout, as the bug which it is hitting, covered under https://github.com/PyCQA/bandit/issues/996, was fixed earlier today and the fix will presumably get to the next bandit release.

PR #204, opened 2023-04-14
Pin bandit version

jm-wk:
We pin the version of Bandit SAST tool to make the CI tests more predictable. If we used the latest version available without pinning it, the CI tests could start failing on new Bandit version without any changes in our codebase.

We pin the Bandit version by adding version requirement to tox.ini.


jm-wk:
I have added a separate commit with workaround for rpm-py-installer. (I think it is better to keep it as a separate commit to allow for easier reverting later on.)

rohanpm (approved):

PR #205, opened 2023-04-17
Resolve Bandit 'request without timeout' findings

jm-wk:
The latest version (1.7.5) of Bandit is reporting requests without timeout in the code as potential vulnerability sources. Since all of the findings are in test code with mock requests, we waive all of them by inline comments.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -52,14 +52,14 @@ deps=
     bandit
 commands=
     # Skip B101 low severity assert warnings in the tests directory
-    bandit -r ./tests --severity-level low  --exit-zero --skip B101
+    bandit -r ./tests --severity-level low --exit-zero --skip B101
     bandit -r . -l --exclude './.tox,./tests' --exit-zero
 
 [testenv:py3-bandit]
 deps=
     bandit
 commands=
-    bandit -r . -ll --exclude './.tox'
+    bandit -r . -ll --exclude './.tox' --confidence-level medium
 
 [pytest]
 testpaths = tests


rohanpm (approved):
I think this is OK, though I was also getting at elsewhere that you could make it run the tool with `--confidence-level medium` which may also fix the problem everywhere, and would avoid similar problems in new code.

jm-wk:
I have reworked the change to ignore the low confidence warnings. 

Should I adjust the confidence levels in other repositories as well, so that the setting is consistent across repositories?

tox.ini
```diff
@@ -52,14 +52,14 @@ deps=
     bandit
 commands=
     # Skip B101 low severity assert warnings in the tests directory
-    bandit -r ./tests --severity-level low  --exit-zero --skip B101
+    bandit -r ./tests --severity-level low --exit-zero --skip B101,B113
```
rohanpm, line 55: I'm not sure why we're adding things to `--skip` here.  It should not affect the correctness of the CI.  Why'd you want to update this?

The original purpose of the two separate testenvs was:

- `bandit` configures the tool to detect only those issues we want CI to enforce
- `bandit-exitzero` configures the tool to find everything it can, but does not enforce the results it finds

It's a valid question whether this is actually working out properly and I wouldn't be surprised if we end up scrapping the exitzero env at some point.  But as long as it exists, I don't get why we'd be adding things into `skip` here as it seems to defeat the original intended purpose.

rohanpm:
CI here has become affected by https://github.com/junaruga/rpm-py-installer/issues/276 and we'll unfortunately need a workaround like https://github.com/release-engineering/pushsource/pull/332 to proceed.

@jm-wk 

> Should I adjust the confidence levels in other repositories as well, so that the setting is consistent across repositories?

You can do that if you want to, but it's definitely scope creep of your current issue.  If I were you, right now I'd only do that in repos where it immediately resolved an issue.

tox.ini
jm-wk: I misunderstood the purpose of the bandit-exitzero testenv and did this to avoid warning fatigue. (I assumed this could be added, since B113 in tests is almost guaranteed to be a false positive.) Changed.

jm-wk:
Thanks for the comments. I have created a new PR with the CI fix here: https://github.com/release-engineering/pubtools-pulplib/pull/206

It seems that working on the issue will take more time than I expected, so I will keep the scope limited and won't attempt to adjust the confidence threshold in all the other repos.

rohanpm:
Could you please rebase this now to get a successful CI run?

rohanpm (approved):

PR #207, opened 2023-05-08
Use rpmdyn in CI rather than rpm-py-installer

rohanpm:
Due to junaruga/rpm-py-installer#276, it was necessary to pin a particular version of virtualenv.

Let's undo that pin and try an alternative provider of RPM bindings instead. This library works a different way which does not require downloading and compiling from RPM sources, and is compatible with the newest versions of pip and other tools.

This is somewhat experimental, so let's try it out here for a while first before rolling out anywhere else.

### DIFF
--- a/.github/workflows/tox-test.yml
+++ b/.github/workflows/tox-test.yml
@@ -14,7 +14,7 @@ jobs:
         with:
           python-version: 3.8
       - name: Install Tox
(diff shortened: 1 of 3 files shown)


rohanpm:
Not currently working due to incompleteness of bindings.

rohanpm:
Now it works.

rbikar (approved):

PR #208, opened 2023-07-18
Add removal function for ud_file_release_mappings_2 [RHELDST-16431]

drepelov:
Before this MR it was possible to only add new mappings to the ud_file_mappings_2 repo note. This meant that if a file was removed from a Pulp repo, the mapping remained in the note. This MR introduces functionality to remove mappings for files that are no longer in that repo.

### DIFF
--- a/pubtools/pulplib/_impl/client/ud_mappings.py
+++ b/pubtools/pulplib/_impl/client/ud_mappings.py
@@ -17,19 +17,20 @@ UD_MAPPINGS_NOTE = os.getenv("PULP_UD_MAPPINGS_NOTE") or "ud_file_release_mappin
 class MappingsHelper(object):
     """A wrapper for the raw release mappings dict.
 
-    This wrapper provides a utility function for adding items to the dict
+    This wrapper provides a utility function for adding and removing items from the dict
     while keeping track of whether any changes have been made.
     """
 
     def __init__(self, data):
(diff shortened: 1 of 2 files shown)


rajulkumar:
Please add the description to the commit as well.
If we revisit this patch in the git log, we'll only the see the header "Add removal function for ud_file_release_mappings_2 [RHELDST-16431]" that won't say much why this patch was added and will have to lookup the jira for details. Also, this being an open sourced library, probably all users won't have access to the jira.

rajulkumar (approved):

pubtools/pulplib/_impl/client/ud_mappings.py
```diff
@@ -164,17 +178,23 @@ def upload_changed_mappings(mappings, repo, repo_url, do_request):
...
     #
     # Returns Future[mappings] once all pages are processed.
 
     for unit in file_page.data:
         version = unit.version
         if version:
+            # Add missing mappings
             mappings.set_file_mapping(version, unit.path, unit.display_order)
+            # Save all the files for which the mapping should exist
+            mappings.filenames_from_pulp.append(unit.path)
```
rohanpm, line 193: I don't think this statement should have been under the conditional.  There are a lot of files in practice which don't have 'version' set and yet shouldn't be removed from the mappings.

PR #209, opened 2023-07-21
Don't overwrite headers passed to Client object

MichalHaluza:
requests module populates Session object with several default headers. The behavior in pubtools-pulpib would cause the default headers dictionary to be completely overwritten by the one passed in kwargs. This is cumbersome in cases where a developer wants to set only a single header, e.g. `User-Agent`. The new behavior ensures the default values are retained and the new values passed in kwargs only update the existing headers dictionary.

### DIFF
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -41,6 +41,14 @@ def test_can_construct_with_session_args(requests_mocker):
     client = Client("https://pulp.example.com/", auth=("x", "y"), verify=False)
 
 
+def test_construct_updates_headers(requests_mocker):
+    """A client instance updates the default requests.Session headers."""
+    client = Client("https://pulp.example.com/", headers={"x": "y"})
+
+    # It should contain the passed header, but it should not be the only header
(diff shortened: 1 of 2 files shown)


rajulkumar (dismissed):

tests/client/test_client.py
```diff
@@ -38,7 +38,7 @@ def test_can_construct_with_throttle_arg(requests_mocker):
 
 def test_can_construct_with_session_args(requests_mocker):
     """A client instance can be constructed with requests.Session kwargs."""
-    client = Client("https://pulp.example.com/", auth=("x", "y"), verify=False)
+    client = Client("https://pulp.example.com/", headers={"x": "y"}, verify=False)
```
rohanpm, line 41: Given that you're implementing special behavior for `headers` argument, a proper test which covers that behavior would be nice.  This change seems the bare minimum to satisfy the coverage tool, only verifying that it doesn't crash.

rohanpm (approved):

PR #211, opened 2023-09-12
Add more attrs backports [RHELDST-20261]

MichalHaluza:
This commit backports `is_callable`, `deep_iterable` and `deep_mapping` methods. These backports are required for pubtools-pulplib to work on RHEL 8 using out-of-the-box versions of python3.6 and python modules, as RHEL 8 comes with ancient attrs 17.4.0

### DIFF
--- a/pubtools/pulplib/_impl/compat_attr.py
+++ b/pubtools/pulplib/_impl/compat_attr.py
@@ -49,10 +49,98 @@ else:
     fields_dict = attr.fields_dict
 
 
+validators = attr.validators
+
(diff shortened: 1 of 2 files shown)


rohanpm (approved):
IMO it'd also be reasonable to just not use these validators or provide no-op implementations on older attr versions, since the validators would usually be expected to find problems during testing and all projects using this should have at least one test configuration using more modern versions.  (The same rationale as how kw_only is handled in the same file.)

rbikar (approved):

PR #212, opened 2023-10-18
Add limit to unassociate [RHELDST-20725]

amcmahon-rh:
We want to perform garbage collection in batches to reduce the amount of resources it consumes and avoid out of memory exceptions. Unit association criteria has a limit field which we could use for batching requests.

### DIFF
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -775,7 +775,7 @@ class Client(object):
             self._do_request, method="POST", url=url, json=body
         )
 
-    def _do_unassociate(self, repo_id, criteria=None):
+    def _do_unassociate(self, repo_id, criteria=None, limit=None):
         url = os.path.join(
             self._url, "pulp/api/v2/repositories/%s/actions/unassociate/" % repo_id
         )
@@ -799,6 +799,9 @@ class Client(object):
                 )
             else:
                 body["criteria"]["filters"] = {"unit": pulp_search.filters}
+
+        if limit:
+            body["criteria"]["limit"] = limit
 
         LOG.debug("Submitting %s unassociate: %s", url, body)
 
--- a/pubtools/pulplib/_impl/model/repository/base.py
+++ b/pubtools/pulplib/_impl/model/repository/base.py
@@ -652,6 +652,10 @@ class Repository(PulpObject, Deletable):
                 be removed.
                 If criteria is omitted, all the content will be removed.
 
+            limit (None, int)
+                Limit the maximum number of units that will be disassociated by
+                pulp.
+
         Returns:
             Future[list[:class:`~pubtools.pulplib.Task`]]
                 A future which is resolved when content has been removed.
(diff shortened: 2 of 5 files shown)


tests/fake/test_fake_remove_content.py
```diff
@@ -30,11 +30,42 @@ def test_can_remove_empty():
...
+    # It should have completed successfully
+    assert task.completed
+    assert task.succeeded
+
+    # It should have removed (only) RPM units
+    assert len(task.units) == 3
+
+
+def test_can_remove_content():
+    """repo.remove() can remove a limited number of units."""
```
rohanpm, line 65: It seems like the doc string between this test and the one above have been swapped around.  Can you check?

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -696,7 +696,13 @@ def remove_content(self, criteria=None, **kwargs):
                     Matcher.in_(type_ids),  # Criteria.with_field_in is deprecated
                 )
 
-        return f_proxy(self._client._do_unassociate(self.id, criteria=criteria))
+        return f_proxy(
+            self._client._do_unassociate(
+                self.id,
+                criteria=criteria,
+                limit=kwargs.get("limit"),
```
rohanpm, line 703: Please document the new argument in this method's doc string.

rohanpm (changes requested):
Not sure what's happened with the CI failure, I'll look at it more thoroughly if it happens again.

pubtools/pulplib/_impl/client/client.py
```diff
@@ -775,7 +775,7 @@ def _do_associate(self, src_repo_id, dest_repo_id, criteria=None, raw_options=No
             self._do_request, method="POST", url=url, json=body
         )
 
-    def _do_unassociate(self, repo_id, criteria=None):
+    def _do_unassociate(self, repo_id, criteria=None, limit=None):
```
rbikar, line 778: After some discussion on private chat -> I think you will likely need also provide `skip` parameter in order to make unassociate to work with units outside the limit (aka first 100 units that match the criteria).

amcmahon-rh: I've opened a WIP PR for pubtools-pulp to show my general idea. It would continue sending unassociate requests sequentially until the number of units removed by a request is less than the limit. As we were discussing, this depends on  implementation detailsof both the python and pulp tasks

rohanpm: How would `skip` be used?

I think it's not needed and the way Aidan is doing it is right, just loop and ask for a certain amount to be removed until there's none left.

If the idea is that `skip` would be used to submit multiple remove tasks concurrently, the naive approach to that wouldn't work.  limit=X skip=Y means "skip the first Y elements and then remove up to X elements".  If you were to queue e.g. limit=1000 skip=0, and limit=1000 skip=1000, Pulp will still process them sequentially, and it's not going to delete the first 2000 elements; it'll delete the first 1000, then leave a gap of 1000 NOT deleted, then delete the next 1000.  Just like how deleting elements from a list while iterating over it often causes similar bugs.

rbikar (approved):

rohanpm (approved):

PR #213, opened 2023-10-19
Don't remove ISOs w/o version from mappings [RHELDST-21014]

MichalHaluza:
Recent commit 92399ba introduced code for removal of deleted ISOs from `ud_file_release_mappings_2` repo note. The line of code collecting ISO filenames present in the repo has been put under the conditional on "version", however not every file has this field populated, which resulted in loss of mapping for all the files where the filed wasn't set. This commit amends that code, so the ISO filename is collected regardless of whether it has the version set.


rbikar (approved):
not sure what's exactly wrong with the pidiff check

rohanpm:
> not sure what's exactly wrong with the pidiff check

There was a backwards-incompatible update to `astroid` library which broke it.  Fixed now.

rohanpm (approved):

PR #217, opened 2023-12-06
Fix the wrong validator of ModulemdUnit.profiles [RHELDST-22116]

dichn:
Currently, the wrong validator resulted in a TypeError, saying pubtools-pulplib is validating the after-conversion "profiles" field as a dict:

TypeError: "'profiles' must be <class 'dict'>
(got frozendict.frozendict({'test': ['secondRepoRPM']}) that is a <class 'frozendict.frozendict'>)."

The validator of ModulemdUnit.profiles should be a frozendict rather than a dict after the "converter=frozenlist_or_none_converter" conversion.

### DIFF
--- a/pubtools/pulplib/_impl/model/unit/modulemd.py
+++ b/pubtools/pulplib/_impl/model/unit/modulemd.py
@@ -9,7 +9,7 @@ from ..convert import (
     frozenlist_or_none_sorted_converter,
     frozendict_or_none_converter,
 )
-from ..validate import optional_list_of, optional_dict
+from ..validate import optional_list_of, optional_frozendict
 
 
 @attr.s(kw_only=True, frozen=True)
@@ -136,7 +136,7 @@ class ModulemdUnit(Unit):
         type=dict,
         pulp_field="profiles",
         default=None,
-        validator=optional_dict,
+        validator=optional_frozendict,
         converter=frozendict_or_none_converter,
     )
     """The profiles of this modulemd unit."""


rohanpm:
There seems to be a deeper story behind this issue.  I find it hard to believe that the bug could be as described because it hasn't changed recently and how could it possibly not have broken anything until now?

Furthermore, these validators are documented as using "isinstance", so when we say optional_dict we're not saying the value has to be *precisely* a dict.  It's allowed to be any subclass of dict, and I would expect frozendict to be a subclass of dict, so it should have been working fine.

Could you check the following in whatever environment you could reproduce this?  For me, frozendict *is* an instance of dict so there is no problem with the validator.

```
>>> from frozendict.core import frozendict
>>> isinstance({}, dict)
True
>>> isinstance(frozendict({}), dict)
True
```

rohanpm:
...and I do have a vague recollection that `frozendict` has both a C and a pure-python implementation, and if you end up using the C implementation it's not a subclass of dict, while if you use the pure-python implementation it is.

dichn:
> Could you check the following in whatever environment you could reproduce this? For me, frozendict is an instance of dict so there is no problem with the validator.

@rohanpm Sure. And it approved your assumption that environment is the cause.
```
I am running on Python 3.9.17 with frozendict 2.3.10
>>> from frozendict.core import frozendict
>>> isinstance({}, dict)
True
>>> isinstance(frozendict({}), dict)
False
```

rohanpm (approved):
I suppose we should still submit this despite that an issue outside of this library triggers the problem.  It should work fine as long as we ensure that the converter and the validator are using the same type (i.e. either both using the pure-python frozendict or both using the C extension frozendict).

PR #218, opened 2024-03-13
Add Client.create_repository() [RHELDST-22483]

rbikar:
`Client` is now capable of creating a repository on pulp
server. The repository is initialized with proper `Importer` in order
to enable sync/upload of content to new repositories.

Another notable changes:
* 409 status is not retried in requests as it is used as indicator of
 existing repository.
* Added create_repository() method to FakeClient as well.
* Small changes in the repository.yaml schema as some of the fields
  don't have to be necessarily initialized when creating repo but can be
  updated later.
* Added `Importer` pulp object mapping class and related subclasses
  that encapsulate minimal required fields for this object with valid
  defaults.
* added `"importers": True` to search options for repos
* fixed converter of `notes.signatures` on `Repository` - empty string is now
  converted to empty list, not to list with empty string

### DIFF
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -278,7 +278,7 @@ class Client(object):
                 Each page will contain a collection of
                 :class:`~pubtools.pulplib.Repository` objects.
         """
-        search_options = {"distributors": True}
+        search_options = {"distributors": True, "importers": True}
         return self._search(
             Repository, "repositories", criteria=criteria, search_options=search_options
         )
@@ -1054,3 +1054,80 @@ class Client(object):
         return self._task_executor.submit(
             self._do_request, method="POST", url=url, json=body
         )
+
+    def create_repository(self, repo):
+        """Create a repository with initial data provided in the
+           argument. Importer for repository is automatically associated
+           if available. If repository already exists, warning is logged.
+           After repository has been successfully created or if repository already exists,
+           it is checked if expected and current repository configuration values are
+           correct.
+
+        Args:
+            repo (:class:`~pubtools.pulplib.Repository`)
+                A repository object used for creation.
+
+        Returns:
+            Future[Repository]
+                A future which is resolved with a value of ``Repository`` once the
+                repository has been created.
+
+        .. versionadded:: 2.39.0
+        """
+        url = os.path.join(self._url, "pulp/api/v2/repositories/")
+
+        body = repo._to_data()
+        repo_id = body["id"]
+
+        importer = body.pop("importers", [])
+        body["importer_type_id"] = importer[0]["importer_type_id"] if importer else None
+        body["importer_config"] = importer[0]["config"] if importer else None
+
+        def log_existing_repo(exception):
+            if (
+                getattr(exception, "response", None) is not None
+                and exception.response.status_code == 409
+            ):
+                LOG.warning("Repository %s already exists", repo_id)
+                return None
+
+            raise exception
+
+        def check_repo(repo_on_server):
+            try:
+                assert (
+                    repo_on_server == repo
+                ), "Repo exists on server with unexpected values"
+            except AssertionError:
+                if importer:
+                    for attr in ["type_id", "config"]:
+                        expected = getattr(repo.importer, attr)
+                        current = getattr(repo_on_server.importer, attr)
+                        if expected != current:
+                            LOG.error(
+                                "Repository %s contains wrong importer %s\n"
+                                "\t expected: %s\n"
+                                "\t current: %s\n",
+                                repo_id,
+                                attr,
+                                expected,
+                                current,
+                            )
+                LOG.exception(
+                    "Repository %s exists on server and contains unexpected values",
+                    repo_id,
+                )
+                raise
+
+            return f_return(repo_on_server)
+
+        LOG.debug("Creating repository %s", repo_id)
+        out = self._request_executor.submit(
+            self._do_request, method="POST", url=url, json=body
+        )
+
+        out = f_map(out, error_fn=log_existing_repo)
+        out = f_flat_map(out, lambda _: self.get_repository(repo_id))
+        out = f_flat_map(out, check_repo)
+
+        return f_proxy(out)
--- a/pubtools/pulplib/_impl/model/repository/base.py
+++ b/pubtools/pulplib/_impl/model/repository/base.py
@@ -8,10 +8,11 @@ from attr import validators, asdict
 from frozenlist2 import frozenlist
 from more_executors.futures import f_proxy, f_map, f_flat_map
 
+from frozendict.core import frozendict  # pylint: disable=no-name-in-module
 from .repo_lock import RepoLock
 from ..attr import pulp_attrib, PULP2_FIELD, PULP2_MUTABLE
 from ..common import PulpObject, Deletable, DetachedException
-from ..convert import frozenlist_or_none_converter
+from ..convert import frozenlist_or_none_converter, frozendict_or_none_converter
 from ..distributor import Distributor
 from ...criteria import Criteria, Matcher
 from ...schema import load_schema
@@ -19,6 +20,7 @@ from ... import compat_attr as attr
 from ...hooks import pm
 from ...util import dict_put, lookup, ABSENT
 
+from ..common import schemaless_init
 
 LOG = logging.getLogger("pubtools.pulplib")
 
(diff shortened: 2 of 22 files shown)


pubtools/pulplib/_impl/client/client.py
```diff
@@ -1054,3 +1054,46 @@ def _do_sync(self, repo_id, sync_options):
...
+        importer = body.pop("importers", [])
+        body["importer_type_id"] = importer[0]["importer_type_id"] if importer else None
+        body["importer_config"] = importer[0]["config"] if importer else None
+
+        def log_existing_repo(exception):
+            if (
+                getattr(exception, "response", None) is not None
+                and exception.response.status_code == 409
+            ):
+                LOG.warning("Repository %s already exists", repo_id)
```
rohanpm, line 1088: I feel a bit uncomfortable that we swallow these errors completely.  It seems like a trap, because someone can do this:

- write a script which creates a repo accidentally setting the wrong config, run it against a server
- realize their mistake, fix the config
- run it again
- their script "succeeds" but it output this warning and they didn't notice.  They believe their script is working OK but actually the importer config is permanently incorrect.

I understand that some handling of 409 would be needed to make the API able to be retried.

What do you think about extending this API so it'd check the repo before returning?
I mean keeping this 409 handling as it is, but at the end of the method it could do:

- fetch the newly created repo
- compare against the one passed in by the user
- if the user's passed Repository set "importer", then compare against the importer actually on the repo, and raise an error if they differ
- the fetched Repository object can be returned (return type `Future[Repository]`)

I believe something like that might be a good idea anyway, because the `Repository` object that the user passed in here is not connected to the client so most of the methods on it can't be used.  The client in most cases probably wants to do something with the `Repository` object, so if this API doesn't fetch and return it they'll probably have to do it themselves afterward.  If they're going to do it anyway, we may as well do it for them which gives us a chance to check if the repo is really what was requested.

pubtools/pulplib/_impl/model/repository/base.py
```diff
@@ -35,6 +36,39 @@ def decorate(klass):
...
+    """
+    Importer is a pulp object that needs to be associated with repository
+    in order to successfully sync or upload content to it.
+    """
+
+    type_id = pulp_attrib(default=None, type=str, pulp_field="importer_type_id")
+    """
+    Type id of the importer.
+    """
+    config = pulp_attrib(default=None, type=dict, pulp_field="config")
```
rohanpm, line 50: Could this please use frozendict (`converter=frozendict_or_none_converter`) so the object is immutable?

pubtools/pulplib/_impl/model/repository/file.py
```diff
@@ -13,6 +13,16 @@
 LOG = logging.getLogger("pubtools.pulplib")
 
 
+@attr.s(kw_only=True, frozen=True)
+class IsoImporter(Importer):
```
rohanpm, line 17: Could we please name this `FileImporter` instead, to match `FileRepository`?

It's the same reasoning as for `FileRepository` itself.  Yes, technically in Pulp's code and type IDs it refers to it as "iso", but that's a mistake and it's actually a generic file handler.  We don't need to repeat the mistake in our own APIs, and by avoiding that we'll be better off if we ever want to port this code to Pulp3.

tests/repository/test_yum_repository.py
```diff
@@ -193,3 +213,72 @@ def test_related_repositories_detached_client():
 
     with pytest.raises(DetachedException):
         repo_binary_test.get_binary_repository()
+
+
+def test_create_repository(client, requests_mocker):
+    repo = YumRepository(id="yum_repo")
+    repo.__dict__["_client"] = client
```
rohanpm, line 220: I don't think this line should be here.  It's a hack done from within some tests to forcibly connect a repo object with a client, but the real users of the library aren't supposed to do that, and in this case it's important to test that `client.create_repository(repo)` actually works without doing this hack first.

I believe if you delete this line the test should still pass and it'd be more accurately covering what the user code should look like.

tests/repository/test_yum_repository.py
```diff
@@ -193,3 +213,72 @@ def test_related_repositories_detached_client():
...
+    assert query["id"] == "yum_repo"
+    # check importer data sent in request body that
+    # are automatically added for yum repository
+    assert query["importer_type_id"] == "yum_importer"
+    assert query["importer_config"] is None
+
+
+def test_create_repository_already_exists(client, requests_mocker, caplog):
+    repo = YumRepository(id="yum_repo")
+    repo.__dict__["_client"] = client
```
rohanpm, line 246: Same here, we shouldn't need this hack.

tests/repository/test_yum_repository.py
```diff
@@ -193,3 +213,72 @@ def test_related_repositories_detached_client():
...
+        # are automatically added for yum repository
+        assert query["importer_type_id"] == "yum_importer"
+        assert query["importer_config"] is None
+        # check logged information about existing repository
+        assert "Repository yum_repo already exists" in caplog.text
+
+
+def test_create_repository_raises_exception(client, requests_mocker):
+    repo = YumRepository(id="yum_repo")
+    repo.__dict__["_client"] = client
```
rohanpm, line 275: And here, we shouldn't need this hack.

rohanpm (changes requested):

rbikar:
I have addressed suggestions but for some reason the coverage checks started to time out.

rohanpm (approved):

rbikar:
@rohanpm 
I've tried to play with the tests and timeout they hit, I was able to simulate it also on local env. but I'm still still unsure why the tests sometimes hang there and sometimes they do not. It seems to hang on `waiter.acquire()`
I believe that that code with chaining futures in the `create_repository()` is OK and I think that there might be some problem in pytest that leads to deadlock. Do you have any idea why this happens? 
 
```
...
  File "/home/runner/work/pubtools-pulplib/pubtools-pulplib/.tox/cov/lib/python3.8/site-packages/more_executors/_impl/map.py", line 72, in _delegate_resolved
    result = self._map_fn(result)
  File "/home/runner/work/pubtools-pulplib/pubtools-pulplib/pubtools/pulplib/_impl/client/client.py", line 1100, in get_and_check_repo
    repo_on_server.result() == repo
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/concurrent/futures/_base.py", line 439, in result
    self._condition.wait(timeout)
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/threading.py", line 302, in wait
    waiter.acquire()
````

**Comment by rohanpm** [comment]
> @rohanpm I've tried to play with the tests and timeout they hit, I was able to simulate it also on local env. but I'm still still unsure why the tests sometimes hang there and sometimes they do not. It seems to hang on `waiter.acquire()` I believe that that code with chaining futures in the `create_repository()` is OK and I think that there might be some problem in pytest that leads to deadlock. Do you have any idea why this happens?

I believe there is a bug in the change actually.  I was able to reproduce it too, but I couldn't reproduce it after a change like below:

```diff
diff --git a/pubtools/pulplib/_impl/client/client.py b/pubtools/pulplib/_impl/client/client.py
index 8a0bc27..5e24a85 100644
--- a/pubtools/pulplib/_impl/client/client.py
+++ b/pubtools/pulplib/_impl/client/client.py
@@ -1093,11 +1093,10 @@ class Client(object):
 
             raise exception
 
-        def get_and_check_repo(_):
-            repo_on_server = self.get_repository(repo_id)
+        def check_repo(repo_on_server):
             try:
                 assert (
-                    repo_on_server.result() == repo
+                    repo_on_server == repo
                 ), "Repo exists on server with unexpected values"
             except AssertionError:
                 if importer:
@@ -1120,7 +1119,7 @@ class Client(object):
                 )
                 raise
 
-            return repo_on_server
+            return f_return(repo_on_server)
 
         LOG.debug("Creating repository %s", repo_id)
         out = self._request_executor.submit(
@@ -1128,6 +1127,9 @@ class Client(object):
         )
 
         out = f_map(out, error_fn=log_existing_repo)
-        out = f_flat_map(out, get_and_check_repo)
+        # After supposedly creating the repo, get it...
+        out = f_flat_map(out, lambda _: self.get_repository(repo_id))
+        # ...and check it matches what's needed
+        out = f_flat_map(out, check_repo)
 
         return f_proxy(out)
```

...i.e. changing it into a fully non-blocking style rather than a mixture of blocking and non-blocking.  Do you want to give that a try?

I think the problem with the other approach is:

- when you call `repo_on_server.result()`, this needs to block on completion of the `get_repository` call.
- `get_repository` needs to go through the `RetryExecutor`'s thread.
- but: that thread might be the one currently in the middle of invoking `get_and_check_repo`.  If so: deadlock!

It can happen randomly, or not, depending on how fast each future manages to complete.  Because when you do `out = f_flat_map(out, get_and_check_repo)`, if the future has already completed by that time, then `get_and_check_repo` will be immediately called in the *current* thread (no deadlock), but if it hasn't completed yet then `get_and_check_repo` will later be called from the `RetryExecutor`'s thread (deadlock).

rohanpm (approved):

PR #222, opened 2024-04-17
Implement tests for fake pulp support of RpmUnit files [RHELDST-21555]

negillett:


negillett:
@rbikar, @rohanpm @rajulkumar, if anyone has any advice here, it would be greatly appreciated.
To me, something seems broken if RpmDependency objects are coming into `convert_field_to_pulp` with subfield 'file' (which does happen without the final conditional). However, I also know what I'm attempting here is "wrong" but I'm not yet sure what is "correct".

tests/fake/test_fake_search_content.py
```diff
@@ -280,3 +300,20 @@ def test_search_content_subfields(populated_repo):
...
     ]
+
+
+def test_search_content_files(populated_repo):
+    """
+    search_content using files subfield in attributes that are lists of strings.
+    """
+    crit = Criteria.and_(
+        Criteria.with_unit_type(RpmUnit),
+        Criteria.with_field("files.file", Matcher.in_(["/some/file"])),
```
rohanpm, line 311: I'm not entirely sure what problem this commit is trying to solve, but I think there are two issues with this line.

Firstly this should probably be "files" and not "files.file".  It is preferred to use the field names on the model classes, the library user is supposed to not necessarily know the details of how the field is actually stored on pulp units.  https://release-engineering.github.io/pubtools-pulplib/api/searching.html#pubtools.pulplib.Criteria.with_field is mentioning this when it says "Supported field names include both model fields and Pulp fields".  The Pulp field names might technically work in some cases but have some limitations, particularly with the fake pulp.

Secondly, the criteria would not be expected to match that RPM.  `Matcher.in_([list of values])` means "value matches some element of this list".  The value on the RPM for the files field is `["/some/file"]` and that value is not contained within the list you passed to `in_`.  Instead, that value *is* the list you passed to `in_`.

It's just like the python operator 'in', e.g. look at this:

```
# the value which is on the rpm.files field
>>> value = ["/some/file"]

# is that value "in" ["/some/file"] ?  No
>>> value in ["/some/file"]
False
```

I don't think the "in" operator is what you'd want for this query at all.  You're not trying to ask "Is the rpm.files value contained in this list of values?", you're trying to ask "Does the rpm.files list contain this value?"

I think when you provide a string value, and mongo finds that the specified field is actually a list, it automatically interprets that as "does the list contain the value".  For instance this command seems to work
to find an RPM containing that filename:

```
pa_tool --env prod content search \
  --content_type rpm \
  --criteria '{"limit": 1, "filters": {"files.file": "/etc/php.d/ncurses.ini"}}'
```

So in summary:

1. I think this should actually be:

```
Criteria.with_field("files", "/some/file")
```

2. Check if the above works against a real Pulp server

3. If it does, then do what's needed to make it work against the fake as well, probably implementing the same kind of "if the target field is a list then check for membership within the list" the same way that mongo apparently does.

negillett: Fixing the criteria alone was enough to get it working against fake pulp.
I'll make the same change in https://github.com/release-engineering/ubi-manifest/pull/136 and see how it goes.
This commit is still useful for covering the file case.

negillett:
cov passed but failed to upload to codecov.io

tests/fake/test_fake_search_content.py
```diff
@@ -32,6 +32,9 @@ def populated_repo(controller):
             sourcerpm="glibc-5.0-1.el5_11.1.src.rpm",
             provides=[RpmDependency(name="gcc")],
         ),
+        RpmUnit(
+            name="jq", version="4.0", release="1", arch="x86_64", files=["/some/file"]
```
rohanpm, line 36: Personally I think it would be a good idea to make the files list contain more than one entry.  There was already some confusion while working on this PR about how searching for a value within a list field works.  Having the list contain only a single value raises the possibility that maybe the code or the test is *only* working in the case where the list has a single value, which is not realistic.

In other words, maybe there is a bug where `Criteria.with_field("files", "/some/file")` only matches an RpmUnit having files equal to *exactly* `["/some/file"]`, but the tests wouldn't be able to expose that bug right now.

rohanpm (approved):

tests/fake/test_fake_search_content.py
negillett: Yeah, good idea

negillett:
Again, just failing upload to codecov

rajulkumar:
> Again, just failing upload to codecov

@negillett https://github.com/release-engineering/pubtools-pulplib/pull/223 should fix the upload

rohanpm (approved):

rajulkumar (approved):

PR #225, opened 2024-09-25
Added support for Distributor when creating repository

rbikar:
Previously it wasn't possible to create repository with distributor because of inconsintent naming of various fields which resulted in Bad requests.

This is now fixed and distributors can be created along repository. Only mininal configuration is currently supported.

### DIFF
--- a/src/pubtools/pulplib/_impl/client/client.py
+++ b/src/pubtools/pulplib/_impl/client/client.py
@@ -10,7 +10,7 @@ import requests
 from more_executors import Executors
 from more_executors.futures import f_map, f_flat_map, f_return, f_proxy, f_sequence
 from io import StringIO
-
+from ..compat_attr import evolve
 from ..model.repository.repo_lock import LOCK_CLAIM_STR
 from ..page import Page
 from ..criteria import Criteria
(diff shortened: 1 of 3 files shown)


src/pubtools/pulplib/_impl/client/client.py
```diff
@@ -1094,25 +1101,37 @@ def log_existing_repo(exception):
...
-                                expected,
-                                current,
+                    self._check_unit(
+                        "importer",
+                        ["type_id", "config"],
+                        repo_updated.importer,
+                        repo_on_server.importer,
+                        repo_updated.id,
+                    )
+                for expected_item in repo_updated.distributors:
```
rajulkumar, line 1124: This will log only the difference in the fields of a specific distributor type e.g. the difference in the fields for say cdn_distributor. 
However, there won't by any log if there are different set of distributors in the repos e.g. repo_on_server has yum and cdn distributors but the incoming or repo_updated has iso and cdn distributors. Only cdn distributors will be compared an logged but difference in presence of iso and yum distributors will be ignored while logging.

rajulkumar (changes requested):

src/pubtools/pulplib/_impl/client/client.py
```diff
@@ -1131,3 +1150,20 @@ def check_repo(repo_on_server):
         out = f_flat_map(out, check_repo)
 
         return f_proxy(out)
+
+    @staticmethod
+    def _check_unit(unit_type, fields, expected_unit, current_unit, repo_id):
```
rohanpm, line 1155: The method and variable naming here seems a bit off to me.  I think `unit` has a specific meaning in Pulp, being a mongo document representing a single piece of content.  I don't think the other types of objects like repositories, distributors, importers are considered "units".  May I suggest something like "resource" might be better?

src/pubtools/pulplib/_impl/client/client.py
rbikar: added logging of missing and extra distributors, not super detailed but i guess good enough to see what's wrong.

rajulkumar (approved):

PR #226, opened 2024-09-26
Populate optional comps.xml unit fields [RHELDST-25995]

amcmahon-rh:
When pushing comps.xml files, if a unit already exists in Pulp, fields that have been removed from the comps.xml did not get overwritten. This resulted in unexpected older data being published from Pulp.

This change extends the comps.xml parsing by giving an option to fill in all fields in comp unit classes from pulp-rpm. When the data is uploaded into pulp this will overwrite old unwanted data.


src/pubtools/pulplib/_impl/comps.py
```diff
@@ -333,7 +361,7 @@ def EndElementHandler(self, _name):
         self.current_buf = StringIO()
 
 
-def units_for_xml(io):
+def units_for_xml(io, fill_fields=False):
```
rohanpm, line 364: I don't think this change alone is going to achieve anything, because this isn't public API.  There is no way for the user of pubtools-pulplib to cause this value to be `True`, so this is just adding a new code path which can never be reached.

The only public API for uploading comps.xml is `YumRepository.upload_comps_xml` method, so if you *wanted* to make the behavior optional then you'd have to update that method to support passing through the new option.

However, I'm not sure there is a good reason to make it optional anyway.  If `YumRepository.upload_comps_xml` supported both cases, it's unlikely most developers would understand how to pick which one is the right behavior for them.  If it had a default option they would most likely not even *try* to understand it, and would just assume that the default option is the appropriate one.

So:
- I definitely think the default behavior needs to be the behavior which will fix our current issue, i.e. needs to be `fill_fields=True`.
- I'd prefer that to be the _only_ behavior, and not to have multiple code paths here at all.

rohanpm (changes requested):

rohanpm (approved):

PR #233, opened 2025-04-01
Temporary fix for `ubi_config_version` attr

rbikar:
* the default value of `ubi_config_version` set to None may cause error when repository is updated on pulp server because pulp  would try to remove non-existent repo note which will fail with KeyError
* let's set this repo note default value to empty string temporarily as the fix should actually be done on pulp
* can/should be reverted as soon as pulp fix is deployed


rbikar:
@rajulkumar  This is pretty pressing issue, I want to unblock rel-engs asap. Merging.

PR #239, opened 2026-06-01
Add pulp3 client [RHELDST-38334]

rbikar:
Added 'raw' client for pulp3 API for basic operations required for:
* repository creation, content search (limited), creation of publication
  and distribution and some other functions like polling tasks.
* written as async using anyio with trio backend, using httpx requests and
  tenacity for retries
* it is expected that this code will undergo further optimization and
  refactoring with higher level of abstraction

### DIFF
--- a/src/pubtools/pulplib/_impl/client_pulp3/client.py
+++ b/src/pubtools/pulplib/_impl/client_pulp3/client.py
@@ -1,0 +1,411 @@
+# Generated-by: Claude Code
+
+from json import JSONDecodeError
+import logging
+import os
+from typing import Any, Dict, Optional
+from urllib.parse import urljoin
+
+import anyio
+import httpx
+from .errors import Pulp3TaskException
+from .retry import get_retry_policy
+
+LOG = logging.getLogger("pubtools.pulplib")
+
+
+class Pulp3Client:
+    """An async client for the Pulp 3.x API.
+
+    This class provides high-level async methods for querying a Pulp 3 server
+    and performing actions in Pulp.
+
+    **Usage:**
+
+    This client is designed to be used as an async context manager with AnyIO's
+    Trio backend:
+
+    .. code-block:: python
+
+        async with Pulp3Client(url="https://pulp.example.com/") as client:
+            repo = await client.create_repository("test-repo")
+            print(repo["name"])
+
+    **Backend:**
+
+    The client uses:
+    - AnyIO with Trio backend for structured concurrency
+    - httpx for async HTTP requests
+    - Task groups for managing concurrent operations
+
+    **Rate Limiting:**
+
+    The client supports automatic rate limiting to avoid overwhelming the server.
+    """
+
+    _DEFAULT_TIMEOUT = 120.0
+    _MAX_CONCURRENT_REQUESTS = int(
+        os.environ.get("PUBTOOLS_PULPLIB_MAX_CONCURRENT_REQUESTS", "10")
+    )
+
+    def __init__(
+        self,
+        url: str,
+        auth: Optional[tuple] = None,
+        cert: Optional[str] = None,
+        verify: bool = True,
+        timeout: Optional[float] = None,
+        max_retries: int = 5,
+        retry_min_wait: float = 1.0,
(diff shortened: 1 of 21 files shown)


rbikar:
pidiff doesn't seem to work properly, but I haven't looked into details

crungehottman:
> pidiff doesn't seem to work properly, but I haven't looked into details

That's my fault, sorry for the confusion. When I created the v2.43.3 release (https://github.com/release-engineering/pubtools-pulplib/commit/247a9bcbe631243586525e16c6de60c3faf0e2f6), I forgot to include the release links in the release commit, which appears to have broken pdiff. I updated the release links in a follow-up commit (https://github.com/release-engineering/pubtools-pulplib/commit/fbfffd9b4f22c382c5846b415167f06b4abe9ce7), so pdiff should pass if you rebase on the latest master.

src/pubtools/pulplib/_impl/client_pulp3/client.py
```diff
@@ -0,0 +1,411 @@
...
+    _DEFAULT_TIMEOUT = 120.0
+    _MAX_CONCURRENT_REQUESTS = int(
+        os.environ.get("PUBTOOLS_PULPLIB_MAX_CONCURRENT_REQUESTS", "10")
+    )
+
+    def __init__(
+        self,
+        url: str,
+        auth: Optional[tuple] = None,
+        cert: Optional[str] = None,
```
crungehottman:
```suggestion
        cert: Optional[Union[str, Tuple[str, str]]] = None,
```

It looks like cert's typehint isn't precise, per https://github.com/encode/httpx/blob/master/httpx/_types.py#L60. 

The pulp3 on Konflux doc requires configuring pulp3 authentication via separate cert and key secrets (and thus separate files) rather than a combined pem file, which would require passing a tuple to the cert parameter.

(auth, verify, and timeout's typehints seem to be imprecise too, but I doubt we'll use the alternatives)

src/pubtools/pulplib/_impl/client_pulp3/retry.py
```diff
@@ -0,0 +1,58 @@
...
+    Args:
+        exception: Exception that occurred
+
+    Returns:
+        True if request should be retried
+    """
+    if isinstance(exception, httpx.HTTPStatusError):
+        status_code = exception.response.status_code
+        # Retry on timeout, rate limiting, and server errors
+        return status_code in (408, 429, 500, 502, 503, 504, 400)
```
crungehottman:
```suggestion
        return status_code in (408, 429, 500, 502, 503, 504)
```

Should HTTP 400 (Bad Request) be retried? Doesn't it usually indicate there's a client side error (like a malformed payload), which wouldn't be fixed by retrying?

tests/client_pulp3/test_client_errors.py
```diff
@@ -0,0 +1,120 @@
...
+
+    async with pulp3_client(max_retries=2) as client:
+        result = await client._request("GET", "/nonexistent/")
+        assert result is None
+        # Should have tried twice (500 triggers retry, then 404)
+        assert route.call_count == 2
+
+
+@pytest.mark.anyio
+async def test_http_400_error_raises(respx_mock, pulp3_client):
```
crungehottman: Related to my earlier comment regarding the list of HTTP status codes that should be retried, this invalid/bad request would be retried up to 5 times before raising.

src/pubtools/pulplib/_impl/client_pulp3/retry.py
rbikar: Yeah, I guess I forgot that while testing retries on real server.
We don't need to retry it.

src/pubtools/pulplib/_impl/client_pulp3/client.py
rbikar: I will run mypy to check this. It's not in CI, not sure if we want to add it though.

rbikar:
@crungehottman 

all should be addressed now (mypy seems to be too strict for sth so not addressing everything now), not sure about pidiff though even with rebase...

rbikar:
self-approved :)

